PR #9, opened 2024-03-11
test: Add mocking

andreaangiolillo:
## Proposed changes

This PR adds mocking to the tools/cli

## Checklist


tools/cli/internal/cli/merge/merge_test.go
```diff
@@ -12,15 +12,52 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build unit
```
gssbzn, line 15: if we don't plan to have int or e2e I think we can skip this for now

tools/cli/internal/cli/merge/merge.go
```diff
@@ -29,14 +29,18 @@ const (
 	DefaultOutputFileName = "FOAS.json"
 )
 
+// writeToFileFunc is a variable that holds the function used to write files.
+// It can be overridden in tests.
+var writeToFileFunc = os.WriteFile
```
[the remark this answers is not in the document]
andreaangiolillo: yes, it is. I was curious to know what you think about this approach 🤣 yes, using afero seems a better approach indeed. I will update it

tools/cli/internal/openapi/openapi.go
```diff
@@ -1,13 +1,71 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package openapi
 
+//go:generate mockgen -destination=../mocks/mock_openapi.go -package=mocks github.com/mongodb/openapi/tools/cli/internal/openapi Parser,Merger
```
gssbzn, line 17: let's avoid some of the CLI mistakes by creating "common" folders and follow a better collocation pattern, let's put the mock next to the file it mocks

tools/cli/internal/openapi/errors/path_conflict_error.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2024 MongoDB Inc
```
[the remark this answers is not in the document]
gssbzn: for apache should be both

tools/cli/internal/cli/merge/merge_test.go
andreaangiolillo: isn't this tag also used to exclude files from the final build when running `go build`?

tools/cli/internal/cli/merge/merge.go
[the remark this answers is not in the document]
gssbzn: use afero or use an interface, I think `io.Writer` is the common file writer interface depends how much magic you wnat since afero will give you a full wrapper on the file system which is handy

tools/cli/internal/cli/merge/merge_test.go
gssbzn: no, tests are excluded by default by the compiler, this was a "clever hack" I did for the CLI to be able to run different tests with different commands, nothing less nothing more

tools/cli/internal/cli/merge/merge.go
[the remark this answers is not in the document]
andreaangiolillo: added afero

tools/cli/Makefile
```diff
@@ -24,15 +24,16 @@ deps:  ## Download go module dependencies
...
 
 .PHONY: setup
 setup: deps devtools ## Set up dev env
 
 .PHONY: fmt
 fmt: ### Format all go files with goimports and gofmt
-	find . -name "*.go" -exec gofmt -w "{}" \;
-	find . -name "*.go" -exec goimports -l -w "{}" \;
+	find . -name "*.go" -not -path "./internal/mocks" -exec gofmt -w "{}" \;
+	find . -name "*.go" -not -path "./internal/mocks" -exec goimports -l -w "{}" \;
```
gssbzn, line 36:
```suggestion
	find . -name "*.go" -exec gofmt -w "{}" \;
	find . -name "*.go" -exec goimports -l -w "{}" \;
```

gssbzn (approved):
:shipit:

PR #34, opened 2024-05-09
feat: default to standard output for the merge command

andreaangiolillo:
## Proposed changes
This PR updates the merge command to default to stdout

### DIFF
--- a/tools/cli/internal/cli/merge/merge.go
+++ b/tools/cli/internal/cli/merge/merge.go
@@ -72,7 +72,7 @@ func prettyPrint(data []byte) error {
 	if err := json.Indent(&prettyJSON, data, "", "    "); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


tools/cli/internal/cli/merge/merge.go
```diff
@@ -65,6 +67,15 @@ func (o *Opts) PreRunE(_ []string) error {
 	return err
 }
 
+func prettyPrint(data []byte) error {
```
jeroenvervaeke, line 70: NIT: Not sure about the name of this function. What about `prettyPrintJson` and `jsonBytes []byte` as argument?

jeroenvervaeke (approved):

tools/cli/internal/cli/merge/merge.go
andreaangiolillo: updated

PR #52, opened 2024-06-21
CLOUDP-257200: Add support to generate an yaml FOAS

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-257200

- add the option to generate FOAS in YAML 

## Checklist


tools/cli/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/stretchr/testify v1.9.0
 	github.com/tufin/oasdiff v1.10.17
 	go.uber.org/mock v0.4.0
+	gopkg.in/yaml.v3 v3.0.1
```
andreaangiolillo, line 12: is there a reason why we want to add support for yaml? The reason why we decided to support only JSON was for some limitations with the tools that we use. I don't remember now in detail, but I think that the external references may not work very for YAML. I need to look at the investigation tickets but so far I've found only this [note](https://docs.google.com/document/d/1gQbM_Wv_hMBospC__iR69sG-c2zPsL5YbCvI9esbyss/edit#bookmark=id.wxhrt6xcvr7e)

blva: we default to json indeed, but some tools may need yaml (SDK gen tools use yaml), so we can just ensure to always generate both but defualt to json

blva: we preferred json because of the versioning annotations and date handling AFAIK

andreaangiolillo (approved):
LGTM

PR #72, opened 2024-07-15
CLOUDP-260459: Fix many diffs between original spec and generated spec

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-260459

- We are only missing 2 types of diff:
 - 1. x-sunset removal, that can only be done after 
 - 2. removal of versions that don't exist in the past (our openapi spec is wrong w/ 2023-01-01 exposing endponits that don't eixst) - see https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/2023-01-01/#tag/Federated-Authentication/operation/createIdentityProvider

## Checklist


tools/cli/internal/openapi/filter/info.go
```diff
@@ -0,0 +1,51 @@
...
+func replaceVersion(inputString string, v *apiversion.APIVersion) string {
+	matches := apiversion.ContentPattern.FindStringSubmatch(inputString)
+	if matches == nil {
+		return inputString // No match found, return the original string
+	}
+
+	replacementString := fmt.Sprintf("application/vnd.atlas.%s+%s", v.String(), matches[4])
+
+	resultString := apiversion.ContentPattern.ReplaceAllString(inputString, replacementString)
+	return resultString
```
gssbzn, line 50:
```suggestion
	replacement := fmt.Sprintf("application/vnd.atlas.%s+%s", v.String(), matches[4])

	return apiversion.ContentPattern.ReplaceAllString(inputString, replacement)
```

tools/cli/internal/openapi/filter/info.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	if info.Description != "" {
+		info.Description = replaceVersion(info.Description, metadata.targetVersion)
+	}
+
+	return nil
+}
+
+func replaceVersion(inputString string, v *apiversion.APIVersion) string {
```
gssbzn, line 41:
```suggestion
func replaceVersion(input string, v *apiversion.APIVersion) string {
```

tools/cli/internal/openapi/filter/info.go
```diff
@@ -0,0 +1,51 @@
...
+	if info == nil {
+		return nil
+	}
+
+	if info.Description != "" {
+		info.Description = replaceVersion(info.Description, metadata.targetVersion)
+	}
+
+	return nil
+}
```
gssbzn, line 39: why the struct if you don't reference `f`? and why the private method hidden behind the public method?

Also I think we need to revisit the linter it should be flagging `f` is unused

blva: you're right, we are initializing the filters without anything in the struct, I think we can initialize them with metadata, WDYT? I can do that in a follow-up PR

gssbzn: @andreaangiolillo seem to be doing something similar on his PR so I'd ask you both to think on the design choices for th elib, do you need a state and as such a struct and how is that struct then initialized and used, and as "state" does that mean I can have multiple cuncurrently?

blva: yeah, we synced today and we think we can make use of keeping metadata in the filters, we initialize them with the right data since each filter should apply to a separate oas

andreaangiolillo (approved):
LGTM

PR #84, opened 2024-07-18
feat: hidden env content type

blva:
## Proposed changes

Add cover for the hiding contents as well

## Checklist


tools/cli/internal/openapi/filter/hidden_envs.go
```diff
@@ -130,6 +147,19 @@ func (f *HiddenEnvsFilter) isResponseHiddenForEnv(response *openapi3.ResponseRef
 	return false
 }
 
+func (f *HiddenEnvsFilter) isContentTypeHiddenForEnv(contentType *openapi3.MediaType) bool {
+	if contentType == nil {
+		return false
+	}
+
+	if extension, ok := contentType.Extensions[hiddenEnvsExtension]; ok {
```
lantoli, line 155: do we need to do the contentType == nil or contentType.Extensions will handle it returning !ok ?

lantoli (approved):

tools/cli/internal/openapi/filter/hidden_envs.go
blva: we do or we'd get a panic on contentType.Extensions, right?

andreaangiolillo (approved):

PR #91, opened 2024-07-19
chore: minor e2e improvements

blva:
## Proposed changes

Just ensuring we're testing different format/files for now. Later we can add more environments.

## Checklist

### DIFF
--- a/tools/cli/test/e2e/cli/split_test.go
+++ b/tools/cli/test/e2e/cli/split_test.go
@@ -16,74 +16,67 @@ var versions = []string{"2023-01-01", "2023-02-01", "2023-10-01", "2023-11-15", 
 
 func TestSplitVersions(t *testing.T) {
 	cliPath := NewBin(t)
+	testCases := []struct {
+		name     string
+		format   string
+		specType string
+		env      string
+	}{
+		{
+			name:     "Split filtered specs json dev",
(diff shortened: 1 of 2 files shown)


tools/cli/test/e2e/cli/split_test.go
```diff
@@ -17,74 +17,69 @@ var versions = []string{"2023-01-01", "2023-02-01", "2023-10-01", "2023-11-15",
 
 func TestSplitVersions(t *testing.T) {
 	cliPath := NewBin(t)
+	testCases := []struct {
```
lantoli, line 20: knit: I recommend to have the test name/description in a map[string]struct instead of one more element in []struct, so it makes it clear that it's not part of the test but the test description itself.

so you can have something like: 
```
...
  "Split filtered specs json dev" : {
			format:   "json",
			specType: "filtered",
			env:      "dev",
},

 "Split filtered specs yaml dev":  {
			format:   "yaml",
			specType: "filtered",
			env:      "dev",
},
...
```

e.g.. https://github.com/mongodb/terraform-provider-mongodbatlas/blob/CLOUDP-261481_cluster_deprecation_banner/internal/service/advancedcluster/model_advanced_cluster_test.go#L157

lantoli (approved):

PR #94, opened 2024-07-22
CLOUDP-261305: Add prod e2e tests

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-261305

## Checklist


tools/cli/test/e2e/cli/split_test.go
```diff
@@ -14,8 +14,8 @@ import (
 )
 
 var versions = []string{
-	"2023-01-01",
-	"2023-02-01",
+	// "2023-01-01",
```
blva, line 17: [need a follow-up PR]

The 2023-01-01 clusters endpoints dont show as deprecated in the current specs, and our logic is marking them as deprecated rightfully. Need to check w/ the team if we want to fix this os ir if was a desired behaviour. 

All other items are matching.

tools/cli/test/e2e/cli/split_test.go
```diff
@@ -113,7 +122,8 @@ func ValidateVersionedSpec(t *testing.T, correctSpecPath, generatedSpecPath stri
 	t.Helper()
 	correctSpec := newOpenAPISpec(t, correctSpecPath)
 	generatedSpec := newOpenAPISpec(t, generatedSpecPath)
-	d, err := diff.Get(diff.NewConfig(), correctSpec, generatedSpec)
+	examples := []string{"examples"}
+	d, err := diff.Get(diff.NewConfig().WithExcludeElements(examples), correctSpec, generatedSpec)
```
blva, line 126: removing examples because our generated spec is matching with the expected examples after we split, but the old ones are not, so I'll ignore for now.

andreaangiolillo: could you add a comment with the explanation so that we don't forget about it?

andreaangiolillo (approved):
LGTM

wtrocki (approved):

tools/cli/test/e2e/cli/split_test.go
blva: yes will add in the follow-up PR!

PR #96, opened 2024-07-23
CLOUDP-263185: Add spectral to the release gh action

andreaangiolillo:
## Proposed changes
CLOUDP-263185

## Description
This PR adds the skeleton of the GH action and adds one step to run spectral. I tested this logic in my fork https://github.com/andreaangiolillo/openapi-test/actions/runs/10056760050


.spectral.yaml
```diff
@@ -0,0 +1,202 @@
+# https://meta.stoplight.io/docs/spectral/cb95cf0d26b83-core-functions
```
blva, line 1: [q] can we have a follow-up task to import these in mms? would be great to have a single point of rules!

.github/workflows/release-spec-prod.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+      - name: Setup Node.js
+        uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b
+        with:
+          node-version: 20
+      - name: Install Spectral
+        run: npm install -g @stoplight/spectral-cli@6.11.1 && spectral --version
```
blva, line 19: [nit] should we make the version a variable? is it even possible? sorry I'm learning XD

.github/workflows/release-spec-prod.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Run Spectral
+        run: spectral lint ./tools/cli/test/data/apiregistry_spec.{json,yml,yaml} --ruleset=.spectral.yaml # we will update this to lint the FOAS in CLOUDP-263186
```
blva, line 21: [nit] can we generate a foas and store there? just so that we can already start checking the full spec! also ok to do in a follow-up

blva: foas using `tools/cli/test/data/openapi-mms.json` - also, can we make sure to capture x-xgen-sha in that file? (I can't open, but I guess its there!)

blva (approved):

.github/workflows/release-spec-prod.yml
andreaangiolillo: yes, I will do it in a follow up PR 👍

.spectral.yaml
andreaangiolillo: sure, it makes sense 👍

.github/workflows/release-spec-prod.yml
andreaangiolillo: doing this in the follow-up PR 👍

.spectral.yaml
andreaangiolillo: https://jira.mongodb.org/browse/CLOUDP-263187

.github/workflows/release-spec-prod.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  run-validations:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+      - name: Setup Node.js
+        uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b
+        with:
+          node-version: 20
```
wtrocki, line 17: When doing global install this step is not required (nodejs is there).
This step gives nice cache capability so if you add spectral to be part of the nodejs package.json (which I strongly recomend) spectral bin will be callable in ./node_modules/bin folder.

Benefits: cuts job runtimes significantly and adds dependabot support.

.github/workflows/release-spec-prod.yml
wtrocki: Made comment above that will enable dependabot support

.github/workflows/release-spec-prod.yml
wtrocki: For maintainability my suggestion would be to always check federated spec as well.
Spectral will be our early canary in the mine for any potential federation issues.

.spectral.yaml
wtrocki: Consider adding Contributing guide. Otherwise upstream PR with new rules can break number of MMS PRs. 
We seeing similar challenges in the MMS validator.

As alternative we can consider backend rules to be source of truth and now (manually) contribute spectral rules upstream.

.github/workflows/release-spec-prod.yml
gssbzn, line 21: did you try https://docs.npmjs.com/cli/v9/commands/npx ? it should save you the install step

andreaangiolillo: no, but I tested it here https://github.com/andreaangiolillo/openapi-test/actions/runs/10057477356/job/27798502158 and it worked! opening a follow up PR

andreaangiolillo: ready for review https://github.com/mongodb/openapi/pull/97

andreaangiolillo: https://github.com/mongodb/openapi/pull/97

.github/workflows/release-spec-prod.yml
andreaangiolillo: https://github.com/mongodb/openapi/pull/97

andreaangiolillo: based on this https://github.com/mongodb/openapi/pull/96#discussion_r1687833169 we won't need the install step

PR #108, opened 2024-07-24
feat: Add flag to skip future versions to split command

andreaangiolillo:
## Proposed changes
This PR adds support to skip future version in the `foascli split` command


tools/cli/internal/cli/split/split.go
```diff
@@ -72,6 +76,20 @@ func (o *Opts) Run() error {
 	return nil
 }
 
+func (o *Opts) skipVersion(version string) bool {
```
blva, line 79: we should expect future version to show up for dev, the hidden-env logic should not generate it if the target is prod and generate if target is dev

andreaangiolillo:
this is not needed as the logic is already there for prod

PR #110, opened 2024-07-24
fix: fixes the filtering logic

andreaangiolillo:
## Proposed changes
The PR proposes the following changes:

- Fixes a bug in the HiddenEnvs filter where we were not deleting the ContentType
- Updates the split command to remove the hidden env before getting the valid version

### DIFF
--- a/tools/cli/internal/openapi/filter/filter.go
+++ b/tools/cli/internal/openapi/filter/filter.go
@@ -14,7 +14,9 @@
 package filter
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/mongodb/openapi/tools/cli/internal/apiversion"
@@ -74,8 +76,30 @@ func initFilters(oas *openapi3.T, metadata *Metadata) error {
 	return nil
 }
 
+func ApplyFiltersWithInit(doc *openapi3.T, metadata *Metadata, init func(oas *openapi3.T, metadata *Metadata) []Filter) error {
+	// make a copy of the oas to avoid modifying the original document when applying filters
+	oas, err := duplicateOas(doc)
+	if err != nil {
+		return err
+	}
+
+	filtersWithInit := init(oas, metadata)
+	for _, filter := range filtersWithInit {
+		if err := filter.Apply(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func ApplyFilters(doc *openapi3.T, metadata *Metadata) error {
-	if err := initFilters(doc, metadata); err != nil {
+	// make a copy of the oas to avoid modifying the original document when applying filters
+	oas, err := duplicateOas(doc)
+	if err != nil {
+		return err
+	}
+
+	if err := initFilters(oas, metadata); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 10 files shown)


tools/cli/internal/openapi/filter/hidden_envs.go
```diff
@@ -109,19 +123,17 @@ func (f *HiddenEnvsFilter) removeResponseIfHiddenForEnv(operation *openapi3.Oper
...
-		for _, contentType := range response.Value.Content {
-			f.removeContentIfHiddenForEnv(contentType)
-		}
-	}
-}
 
-func (f *HiddenEnvsFilter) removeContentIfHiddenForEnv(contentType *openapi3.MediaType) {
-	if isContentTypeHiddenForEnv := isContentTypeHiddenForEnv(contentType, f.metadata.targetEnv); isContentTypeHiddenForEnv {
-		log.Printf("Removing contentType: %q because is hidden for target env: %q", contentType.Schema.Ref, f.metadata.targetEnv)
-		contentType.Schema = nil // Remove ContentType if it is hidden for the target environment
```
andreaangiolillo, line 121: `contentType.Schema = nil // Remove ContentType if it is hidden for the target environment` This piece of code was removing the schema instead of the ContentType

tools/cli/internal/openapi/versions.go
```diff
@@ -18,13 +18,24 @@ import (
...
 // ExtractVersions extracts version strings from an OpenAPI specification.
-func ExtractVersions(oas *openapi3.T) []string {
-	versions := make(map[string]struct{})
+func ExtractVersions(oas *openapi3.T, env string) ([]string, error) {
+	// We need to remove the version that are hidden for the given environment
+	doc, err := filter.ApplyFiltersWithInit(oas, filter.NewMetadata(nil, env), func(oas *openapi3.T, metadata *filter.Metadata) []filter.Filter {
+		return []filter.Filter{
+			filter.InitHiddenEnvsFilter(oas, metadata),
+		}
+	})
```
andreaangiolillo, line 31: Applying the hidden env filter when retrieving the versions for a spec

tools/cli/internal/openapi/filter/hidden_envs.go
andreaangiolillo: The main reason was that we need to pass the `k` and `response.Value.Content` as we need them to delete the content.

tools/cli/internal/openapi/filter/filter.go
```diff
@@ -74,15 +76,54 @@ func initFilters(oas *openapi3.T, metadata *Metadata) error {
 	return nil
 }
 
-func ApplyFilters(doc *openapi3.T, metadata *Metadata) error {
-	if err := initFilters(doc, metadata); err != nil {
-		return err
+func ApplyFiltersWithInit(doc *openapi3.T, metadata *Metadata, init func(oas *openapi3.T, metadata *Metadata) []Filter) (*openapi3.T, error) {
```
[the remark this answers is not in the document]
andreaangiolillo: do you think the problem is the name and/or the fact that we are passing a function returning an array of filters?

I can call the function `newFilters` if we think `init` is the problem

tools/cli/internal/openapi/versions.go
blva: if no env is provided as a flag we should return all versions in the v2 file

blva: the core of the command is to find versioned matching content in the spec. if an env is provided, then we apply filtering

tools/cli/internal/openapi/filter/hidden_envs.go
```diff
@@ -30,6 +30,13 @@ type HiddenEnvsFilter struct {
 	metadata *Metadata
 }
 
+func InitHiddenEnvsFilter(oas *openapi3.T, metadata *Metadata) *HiddenEnvsFilter {
```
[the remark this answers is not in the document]
andreaangiolillo: I removed new Hidden envs. Let me know what you think

tools/cli/internal/openapi/filter/filter.go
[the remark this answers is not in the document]
andreaangiolillo: Updated. Let me know what you think

tools/cli/internal/openapi/versions.go
andreaangiolillo: Updated. Thanks!

blva (approved):
lgtm thanks for the changes!

PR #113, opened 2024-07-25
CLOUDP-263186: Add Logic to retrieve the FOASMetadata from APIx s3 bucket

andreaangiolillo:
This PR adds the logic to retrieve the FOASMetadata from APIx s3 bucket.

I tested my changes in a fork: https://github.com/andreaangiolillo/openapi-test/actions/runs/10096226482

Screenshot 2024-07-25 at 15 46 46

# Next Steps
I plan to move the FOAS generation logic to another workflow that is called from the main one similar to the validations.

### DIFF
--- a/.github/workflows/release-spec-prod.yml
+++ b/.github/workflows/release-spec-prod.yml
@@ -7,20 +7,80 @@ permissions:
   contents: write
   pull-requests: write
 jobs:
+  generata-spec:
+    name: Generate the Federated OpenAPI Spec
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+        with:
+          sparse-checkout: |
+            tools
+      - name: Install Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32
+        with:
+          go-version-file: 'tools/cli/go.mod'
+          cache-dependency-path: "tools/cli/go.sum"
+      - name: Build CLI
+        working-directory: tools/cli
+        run: make build
+      - name: Retrieve Specs
+        working-directory: tools/cli/bin
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-spec-prod.yml
```diff
@@ -7,20 +7,80 @@ permissions:
...
+        run: make build
+      - name: Retrieve Specs
+        working-directory: tools/cli/bin
+        env:
+          AWS_DEFAULT_REGION: ${{vars.AWS_DEFAULT_REGION}}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_PROD }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_PROD }}
+          S3_BUCKET: ${{ vars.S3_BUCKET_PROD }}
+        run: |
+          aws s3 cp s3://"${S3_BUCKET}"/openapi/foas-metadata.json foas-metadata.json
```
blva, line 35: [nit] can this live in a .sh file so that it can be linted?

andreaangiolillo: yes, but it is already linted by the `code_health`worfkflow at the step `actionlint`

andreaangiolillo: happy to move it to a script in a follow up PR if you think would help 👍

blva: based on evergreen life experience, I'd rather have as much logic as possible out of actions and into something we can unit test, and actions would only pass the secrets/variables. that way we can see how our changes affect those flows ahead of time. not a strong opinion, up to you. e.g. build a foascli release command or a releasecli to do the foas metadata handling (think that in the future we can change that format, or s3 bucket structure, or how it works for each branch)

andreaangiolillo: addressed this comment in https://github.com/mongodb/openapi/pull/114

PR #118, opened 2024-07-29
CLOUDP-264326: Compare and send Postman fork metrics

Luke-Sanderson:
## Proposed changes

- Get the previous weeks metrics from cache 
- Compare the fork numbers from each file using jq and collection ids
- Construct a summary of the changes
- Send the message to slack using webhook secret

_Jira ticket:_ CLOUDP-264326

Tested using fake data on fork.

Screenshot 2024-07-29 at 15 29 19
Screenshot 2024-07-29 at 15 30 55

## Checklist


Luke-Sanderson:
Note: I have updated the message to say since last update rather than last week but didn't rerun the manual test on the fork

.github/workflows/postman-metrics.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Fetch previous week's data from cache
+      uses: actions/cache/restore@v4
+      with:
+        path: ./tools/postman/openapi/fork-data.json
+        key: fork-data-${{ github.run_id }}
+        restore-keys: |
+          fork-data-
+
+    - name: Save previous week's data to file
```
andreaangiolillo, line 26: why did we decide to use previous week instead of something more recent?

.github/workflows/postman-metrics.yml
```diff
@@ -0,0 +1,51 @@
...
+      working-directory: ./tools/postman/openapi
+      run: |
+        if [ -f fork-data.json ]; then
+          cp fork-data.json previous-fork-data.json
+        else
```
andreaangiolillo, line 31: I don't understand why you are still creating the file if you don't have data. Could you clarify a bit more here? Can't you just don't compare if there are no data?

.github/workflows/postman-metrics.yml
```diff
@@ -0,0 +1,51 @@
...
+          echo '{"collections":[]}' > previous-fork-data.json
+        fi
+
+    - name: Fetch current fork data
+      working-directory: ./tools/postman
+      run: |
+        make fetch_forks
```
andreaangiolillo, line 38:
```suggestion
      run: make fetch_forks 
```

.github/workflows/postman-metrics.yml
```diff
@@ -0,0 +1,51 @@
...
+      working-directory: ./tools/postman
+      run: |
+        make fetch_forks 
+
+    - name: Compare data.json with previous week's data
+      working-directory: ./tools/postman
+      env: 
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      run: |
+        make compare_and_send_forks
```
andreaangiolillo, line 45:
```suggestion
      run: make compare_and_send_forks
```

.github/workflows/release-postman.yml
```diff
@@ -1,8 +1,9 @@
 name: Postman Release
 on:
   # Enabled by CLOUDP-262512
-  #  schedule:
-  #    - cron: '0 0 10 ? * *'
+  # schedule:
+  #   - cron: '0 10 ? * *'
+  #   - cron: '0 10 ? * 1'
```
andreaangiolillo, line 6: I am not a fan of this approach. Could you split this action into two?

tools/postman/Makefile
```diff
@@ -29,6 +29,10 @@ build: fetch_openapi convert_to_collection
 .PHONY: build_and_upload
 build_and_upload: build upload_collection
 
+.PHONY: compare_and_send_forks
+compare_and_send_forks: 
+	./scripts/send-fork-data.sh
+
```
andreaangiolillo, line 35: Could you split this to have  `make compare` and `make send-data` and then  `make compare_and_send_forks` will call both make commands. 
the reason is that you may want to compare without sending the data

.github/workflows/postman-metrics.yml
Luke-Sanderson: I should have changed the name to previous runs data since it can also be triggered manually. Stephen wanted weekly updates, which is why the cron is set to weekly.

.github/workflows/postman-metrics.yml
Luke-Sanderson: My thought process was that if there is no previous data then it is equivalent to if there was no collections the previous week, which was an edge case I was testing.

I agree that it does look confusing so I will add a check that the file exists in the bash script rather than creating an empty result

Luke-Sanderson: Added check in bash script

tools/postman/scripts/compare-forks.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+set -o errexit
```
andreaangiolillo, line 2:
```suggestion
set -o errexit
```

Luke-Sanderson: I will change this for all scripts in a different PR. Thank you for all the comments!

PR #125, opened 2024-07-31
fix: FOAS Prod GH Action creates wrong issue

andreaangiolillo:
## Proposed changes
This PR fixes a bug with the FOAS GH action that creates multiple issue when one step fails

fixes: #123


.github/workflows/optional-spec-validations.yml
```diff
@@ -41,7 +42,7 @@ jobs:
           npx -- @stoplight/spectral-cli@"${SPECTRAL_VERSION}" lint ./tmp/collection.json --ruleset=./validation/spectral.yaml
           popd
       - name: Create Issue
-        if: ${{ failure() }}
+        if: ${{ failure() && steps.spectral-validation.outcome == 'failure' }}
```
blva, line 45: can you explain a bit how this fix the issue? (newbie)

blva (approved):

.github/workflows/optional-spec-validations.yml
andreaangiolillo: sure

` failure() && steps.spectral-validation.outcome == 'failure'`
` failure()` check if the overall GH action has failed.
`steps.spectral-validation.outcome == 'failure'` makes sure that the step with the id `spectral-validation` has failed 

Only if both conditions are satisfied, the Create Issue step will run.

Before the fix, the only condition to create the issue was that the overall action has failed.

PR #140, opened 2024-08-07
CLOUDP-26613: add normalization to the changelog

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-26613

This PR adds support for the first step of the API Changelog which is normalizing the input specs.

### DIFF
--- a/tools/cli/internal/cli/changelog/create.go
+++ b/tools/cli/internal/cli/changelog/create.go
@@ -43,7 +43,7 @@ func (o *Opts) Run() error {
 		return err
 	}
 
-	fmt.Print("Normalizing the v2.json spec...\n")
+	fmt.Print("Printing the basr v2.json spec normalized...\n")
 	base, err := json.MarshalIndent(*changelog.Base, "", "  ")
 	if err != nil {
 		return err
--- a/tools/cli/internal/cli/changelog/create.go
+++ b/tools/cli/internal/cli/changelog/create.go
@@ -15,20 +15,42 @@
 package changelog
 
 import (
+	"encoding/json"
+	"fmt"
+
 	"github.com/mongodb/openapi/tools/cli/internal/cli/flag"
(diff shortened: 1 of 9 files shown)


tools/cli/internal/cli/changelog/create.go
```diff
@@ -15,20 +15,42 @@
...
-	dryRun       bool
+	fs              afero.Fs
+	basePath        string
+	revisionPath    string
+	exceptionsPaths string
+	dryRun          bool
 }
 
 func (o *Opts) Run() error {
+	changelog, err := openapi.NewChangelog(
```
blva, line 37: [nit] could we clean up test code and add e2e for the command?

tools/cli/internal/openapi/openapi3.go
```diff
@@ -59,6 +61,35 @@ func (o *OpenAPI3) CreateOpenAPISpecFromPath(path string) (*load.SpecInfo, error
...
+	if err != nil {
+		return nil, err
+	}
+	return &load.SpecInfo{
+		Spec: spec,
+	}, nil
+}
+
+// normalizeMediaType replaces versioned media types (e.g. application/vnd.atlas.2023-01-01) with standard media types (e.g. application/json).
+func normalizeMediaType(sourceFile []byte) string {
```
blva, line 82: [q] can we add unit test for this?

tools/cli/internal/cli/changelog/create.go
andreaangiolillo: the command is not doing anything beside the normalization, if we add the e2e test now we will need to change it completely once the command create the actual changelog. I think we can wait for e2e tests since the command is not ready

tools/cli/internal/openapi/openapi3.go
andreaangiolillo: updated

blva (approved):

PR #143, opened 2024-08-07
CLOUDP-266130: Add changelog core logic

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-266130

This PR adds changelog core logic to FOAS CLI

```bash
./bin/foascli changelog create -b ~/Downloads/test/base -r ~/Downloads/test/revision -e ~/Downloads/test/exceptions.yaml                         
Printing the checks
[
  {
    "Attributes": null,
    "Id": "api-schema-removed",
    "Args": [
      "AzureHardwareSpec20250101"
    ],
    "Comment": "",
    "Level": 3,
    "Component": "schemas",
    "SourceFile": "",
    "SourceLine": 0,
    "SourceLineEnd": 0,
    "SourceColumn": 0,
    "SourceColumnEnd": 0
  },
.....
]
```

### DIFF
--- a/tools/cli/internal/openapi/openapi.go
+++ b/tools/cli/internal/openapi/openapi.go
@@ -97,14 +97,14 @@ func NewChangelog(base, revision, exceptionFilePath string) (*Changelog, error) 
 
 	changelogConfig := checker.NewConfig(
 		checker.GetAllChecks()).WithOptionalChecks(breakingChangesAdditionalCheckers).WithDeprecation(deprecationDaysBeta, deprecationDaysStable)
-	
-		return &Changelog{
+
+	return &Changelog{
 		Base:              baseSpec,
 		Revision:          revisionSpec,
 		ExceptionFilePath: exceptionFilePath,
 		Config:            changelogConfig,
 		OasDiff: &OasDiff{
-			base:   baseSpec,
+			base:     baseSpec,
 			external: revisionSpec,
 			config: &diff.Config{
 				IncludePathParams: true,
--- a/tools/cli/internal/openapi/openapi.go
+++ b/tools/cli/internal/openapi/openapi.go
@@ -97,11 +97,19 @@ func NewChangelog(base, revision, exceptionFilePath string) (*Changelog, error) 
 
 	changelogConfig := checker.NewConfig(
 		checker.GetAllChecks()).WithOptionalChecks(breakingChangesAdditionalCheckers).WithDeprecation(deprecationDaysBeta, deprecationDaysStable)
-	return &Changelog{
+	
+		return &Changelog{
 		Base:              baseSpec,
 		Revision:          revisionSpec,
 		ExceptionFilePath: exceptionFilePath,
 		Config:            changelogConfig,
+		OasDiff: &OasDiff{
+			base:   baseSpec,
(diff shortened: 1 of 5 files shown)


tools/cli/internal/openapi/openapi.go
```diff
@@ -97,11 +97,19 @@ func NewChangelog(base, revision, exceptionFilePath string) (*Changelog, error)
...
+
 	return &Changelog{
 		Base:              baseSpec,
 		Revision:          revisionSpec,
 		ExceptionFilePath: exceptionFilePath,
 		Config:            changelogConfig,
+		OasDiff: &OasDiff{
+			base:     baseSpec,
+			external: revisionSpec,
+			config: &diff.Config{
```
blva, line 109: similar to breaking changes, we need to keep the same config so that we classify all changes the same
Reference:
https://github.com/mongodb/openapi/blob/97a837768a48e1b562a9b9be8232126aec05f783/tools/cli/internal/openapi/openapi.go#L99


should we then use changelogConfig here?

tools/cli/internal/openapi/openapi.go
```diff
@@ -97,11 +97,19 @@ func NewChangelog(base, revision, exceptionFilePath string) (*Changelog, error)
 
 	changelogConfig := checker.NewConfig(
 		checker.GetAllChecks()).WithOptionalChecks(breakingChangesAdditionalCheckers).WithDeprecation(deprecationDaysBeta, deprecationDaysStable)
+
 	return &Changelog{
 		Base:              baseSpec,
 		Revision:          revisionSpec,
 		ExceptionFilePath: exceptionFilePath,
 		Config:            changelogConfig,
+		OasDiff: &OasDiff{
```
blva, line 106: we have also been using --flatten flag, can you check if there is a way to define this in oasdiff config? if not, we can create a ticket to investigate if we can use the flattening somehow, and i can check or we can try to contribute, but I don't think it's a blocker for this PR

tools/cli/internal/openapi/openapi.go
andreaangiolillo: those are two different config, the changelog uses `*checker.Config` and diff use `*diff.Config`

blva: oh got it! then we need to add the optional checks to the diff as well if possible (if not feel free to resolve)

tools/cli/internal/openapi/openapi.go
andreaangiolillo: updated

tools/cli/internal/openapi/openapi.go
andreaangiolillo: no, there is no option for optionals checks in the Diff. I am using the same Diff.config we use for the merge command

this is the config for the Diff
```go
/ Config includes various settings to control the diff
type Config struct {
	PathFilter              string
	FilterExtension         string
	PathPrefixBase          string
	PathPrefixRevision      string
	PathStripPrefixBase     string
	PathStripPrefixRevision string
	ExcludeElements         utils.StringSet
	IncludePathParams       bool
}
```

blva (approved):

PR #148, opened 2024-08-08
CLOUDP-266435: add the transformation logic for enum

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-266435


tools/cli/internal/changelog/outputfilter/squash.go
```diff
@@ -150,6 +160,19 @@ func squashEntriesByValues(
 	return result, nil
 }
 
+func replaceOnlyFirstOccurrence(template, valuesToAddToTemplate string) string {
```
blva, line 163: [nit] can we add a comment somewhere on why we need to ensure replacing only the first occurrence?

blva (approved):

tools/cli/internal/changelog/outputfilter/squash.go
andreaangiolillo: sure, I will address this comment in the next PR

PR #150, opened 2024-08-09
CLOUDP-266443: Add support to handle hidden entries in changelog

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-266443

## Checklist


tools/cli/internal/changelog/changelog.go
```diff
@@ -42,10 +42,10 @@ type Metadata struct {
 	Revision          *load.SpecInfo //  the new spec to compare against the base
 	Config            *checker.Config
 	OasDiff           *openapi.OasDiff
-	ExceptionFilePath string
+	ExemptionFilePath string
```
andreaangiolillo, line 45: oh wow I've just noticed that the file name is Exemption not exception 🤦

tools/cli/internal/breakingchanges/exemptions.go
```diff
@@ -57,7 +61,8 @@ func transformComponentEntry(breakingChangeDescription string) string {
 	return breakingChangeDescription
 }
 
-func getValidExemptionsList(exemptionsPath string, ignoreExpiration bool, fs afero.Fs) ([]Exemption, error) {
+// GetValidExemptionsList returns a list of exemptions. If ignoreExpiration is set to true, it will return all exemptions.
+func GetValidExemptionsList(exemptionsPath string, ignoreExpiration bool, fs afero.Fs) ([]Exemption, error) {
```
andreaangiolillo, line 65: I think we could use `os.Open` instead of afero since it doesn't bring any value using it here but we are passing it as input params from different functions

tools/cli/internal/breakingchanges/exemptions.go
```diff
@@ -27,7 +27,11 @@ type Exemption struct {
 	BreakingChangeDescription string `yaml:"breaking_change_description"`
 	ExemptUntil               string `yaml:"exempt_until"`
 	Reason                    string `yaml:"reason"`
-	HideFromChangelog         string `yaml:"hide_from_changelog"`
+	HideFromChangelog         string `yaml:"hide_from_changelog:omitempty"`
```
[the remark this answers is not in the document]
blva: unfortunately it is a string in the current exemptions file, so I'd rather not change it for now

tools/cli/internal/changelog/changelog.go
blva: ahhaha no worries

tools/cli/internal/breakingchanges/exemptions.go
blva: i think it's a pattern i noticed in our commands, but I thinkwe should start changing to afero as we can then inject files in tests much easily and avoid file issues, let me know what you think i can revert back and change all in one pr but i think it will take time

andreaangiolillo (approved):

PR #152, opened 2024-08-09
fix: fixes `branded_preview.sh` to run the version command with the correct branch name for prod + update readme with new badge 

andreaangiolillo:
## Proposed changes
This PR updates the `branded_preview.sh` to run the foascli version command with the correct branch name for prod. It also update the the readme.md with a badge opening the preview file.
Screenshot 2024-08-09 at 16 39 00


.github/scripts/branded_preview.sh
```diff
@@ -1,12 +1,13 @@
 #!/bin/bash
 set -eou pipefail
 
+foascli versions -s v2.json --env "${target_env:?}" -o versions.json
+
```
andreaangiolillo, line 5: main fix, we were providing the env=main instead of prod, defaulting to dev

blva (approved):

PR #156, opened 2024-08-11
CLOUDP-266992: Migrate changes_processor.py logic to foascli

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-266992

### DIFF
--- a/tools/cli/internal/changelog/merge.go
+++ b/tools/cli/internal/changelog/merge.go
@@ -18,9 +18,6 @@ const (
 	changeTypeUpdate        = "update"
 	changeTypeDeprecated    = "deprecate"
 	changeTypeRemove        = "remove"
-	ChangeLevelErr          = 3
-	ChangeLevelWarn         = 2
-	ChangeLevelInfo         = 1
 )
 
 func newChangeTypePriority() map[string]int {
@@ -38,6 +35,7 @@ func newChangeTypeOverrides() map[string]string {
 	}
 }
 
+// MergeChangelog merges the base changelog with the new changes from a Base and Revision OpenAPI specs
 func (m *Metadata) MergeChangelog() ([]*Entry, error) {
 	changes, err := m.newOasDiffEntries()
 	if err != nil {
@@ -52,6 +50,16 @@ func (m *Metadata) MergeChangelog() ([]*Entry, error) {
 	return m.mergeChangelog(changeTypeUpdate, changes, conf)
 }
 
+// mergeChangelog merges the base changelog with the new changes
+// Logic:
+// 1. If the entry already exists in the changelog for the Run Date, use that entry or create it (newEntryAtRunDate)
+// 2. Get only the deprecated by newer version changes (newDeprecatedByNewerVersionChanges)
+// 3. Run newMergedChanges for deprecated changes
+// 4. Get only the revision changes (newRevisionChanges)
+// 5. Run newMergedChanges for revision changes
+// 6. If there are any changes, append them to the entry
+// 7. Append the entry to the changelog
+// 8. Sort the changelog by date DESC, path + httpMethod ASC, version DESC
 func (m *Metadata) mergeChangelog(
 	changeType string,
 	changes []*outputfilter.OasDiffEntry,
@@ -61,12 +69,7 @@ func (m *Metadata) mergeChangelog(
 		return nil, err
 	}
 
-	index, entry := m.newEntryAtRunDate(changelog)
-	if index != -1 {
-		// If the entry already exists, remove it from the changelog
-		changelog = append(changelog[:index], changelog[index+1:]...)
-	}
-
+	entry := m.newEntryAtRunDate(&changelog)
 	depreactedChanges := m.newDeprecatedByNewerVersionChanges(changes, conf)
 	mergedDeprecatedPathsChanges, err := newMergedChanges(depreactedChanges, changeTypeDeprecated, m.Base.Version, entry.Paths, conf)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


tools/cli/internal/changelog/merge.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+}
+
+func newChangeTypeOverrides() map[string]string {
+	return map[string]string{
+		endpointAddedCode: changeTypeRelease,
+	}
+}
+
+func (m *Metadata) MergeChangelog() ([]*Entry, error) {
```
blva, line 41: [nit] document public func

tools/cli/internal/changelog/merge.go
```diff
@@ -0,0 +1,323 @@
...
+	conf map[string]*outputfilter.OperationConfigs) ([]*Entry, error) {
+	changelog, err := duplicateChangelog(m.BaseChangelog)
+	if err != nil {
+		return nil, err
+	}
+
+	index, entry := m.newEntryAtRunDate(changelog)
+	if index != -1 {
+		// If the entry already exists, remove it from the changelog
+		changelog = append(changelog[:index], changelog[index+1:]...)
```
blva, line 67: [q] why? I didn't find this in the code, should we remove or should we merge the already ran ones? is there any risk we could lose data here?

tools/cli/internal/changelog/merge.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	depreactedChanges := m.newDeprecatedByNewerVersionChanges(changes, conf)
+	mergedDeprecatedPathsChanges, err := newMergedChanges(depreactedChanges, changeTypeDeprecated, m.Base.Version, entry.Paths, conf)
+	if err != nil {
+		return nil, err
+	}
+
+	revisionChanges := m.newRevisionChanges(changes)
```
blva, line 76: [nit[ it's a bit difficult to understand what you're doing in `mergeChangelog` at a high levle, could this part from line 76 to 8- be a func?

blva: maybe a ccomment in the func could also help

tools/cli/internal/changelog/merge.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if len(changes) == 0 {
+		return m.BaseChangelog, nil
+	}
+
+	conf := outputfilter.NewOperationConfigs(m.Base, m.Revision)
+	return m.mergeChangelog(changeTypeUpdate, changes, conf)
+}
+
+func (m *Metadata) mergeChangelog(
```
blva, line 55: can we explicitly filter the path changes somewhere here, or we don't need it?

tools/cli/internal/changelog/merge.go
```diff
@@ -0,0 +1,323 @@
...
+const (
+	endpointAddedCode       = "endpoint-added"
+	endpointDeprecatedCode  = "endpoint-deprecated"
+	endpointReactivatedCode = "endpoint-reactivated"
+	notSetPriority          = 10
+	changeTypeRelease       = "release"
+	changeTypeUpdate        = "update"
+	changeTypeDeprecated    = "deprecate"
+	changeTypeRemove        = "remove"
+	ChangeLevelErr          = 3
```
blva, line 21: nit - are we able to import this from oasdiff? (once they changed it and it messed a bunch of stuff)

tools/cli/internal/changelog/merge.go
```diff
@@ -0,0 +1,323 @@
...
+	if index != -1 {
+		// If the entry already exists, remove it from the changelog
+		changelog = append(changelog[:index], changelog[index+1:]...)
+	}
+
+	depreactedChanges := m.newDeprecatedByNewerVersionChanges(changes, conf)
+	mergedDeprecatedPathsChanges, err := newMergedChanges(depreactedChanges, changeTypeDeprecated, m.Base.Version, entry.Paths, conf)
+	if err != nil {
+		return nil, err
+	}
```
blva, line 74: nit: move this into another func if possible, andsay what it does

do we detect deprecation changes and merge them?

tools/cli/internal/changelog/merge.go
andreaangiolillo: This piece of logic is the one inside `_insert_date_if_absent`. In python, we return the same entry we have in the changelog and work on it. With golang if we don't want to remove the same entry, we need to return a pointer to the element in the array from `_insert_date_if_absent`, which means that we need to first pass a pointer to the changelog `*[]*Entry` and then return a *Entry. I will try to use this logic to avoid deletion.

tools/cli/internal/changelog/merge.go
andreaangiolillo: updated

tools/cli/internal/changelog/merge.go
andreaangiolillo: > do we detect deprecation changes and merge them?
yes

updated.

tools/cli/internal/changelog/merge.go
andreaangiolillo: updated

tools/cli/internal/changelog/merge.go
andreaangiolillo: updated

tools/cli/internal/changelog/merge.go
andreaangiolillo: updated

tools/cli/internal/changelog/merge.go
andreaangiolillo: we filter the path changes inside `newPathsFromChanges`

blva (approved):
LGTM, thanks for the changes!!

PR #169, opened 2024-08-15
CLOUDP-266429: Update the changelog create command to compare all the versions

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-266429

This PR adds the changelog versions logic to changelog create command. I tested the generated file with the ones in [evergreen](https://evergreen.mongodb.com/task/mms_v20240731_generate_changelog_GENERATE_CHANGELOG_66b33417bbd01600072e92cb_24_08_07_08_45_11) and they matched

**We have a separate ticket to add the e2e tests**


tools/cli/internal/changelog/changelog.go
```diff
@@ -77,7 +78,7 @@ type Path struct {
 	Tag            string     `json:"tag"`
 	StabilityLevel string     `json:"stabilityLevel,omitempty"`
 	ChangeType     string     `json:"changeType,omitempty"`
-	Changes        []Change   `json:"changes,omitempty"`
+	Changes        []*Change  `json:"changes,omitempty"`
```
jeroenvervaeke, line 81: [q] just for my understanding, why are you using a pointer here?
what's the difference between `nil` and an empty slice?

andreaangiolillo: Because I am passing `changes` to different functions and it avoids copying the value on each of them and it should be more efficient during the append https://stackoverflow.com/questions/27622083/slices-of-structs-vs-slices-of-pointers-to-structs

tools/cli/internal/openapi/file.go
```diff
@@ -31,20 +32,49 @@ const (
...
+		if err != nil {
+			return err
+		}
+
+		path = newPathWithExtension(path, YAML)
+		if err := afero.WriteFile(fs, path, dataYAML, 0o600); err != nil {
+			return err
+		}
+
+		log.Printf("\nnFile was saved in '%s'.\n\n", path)
```
blva, line 64: is \nn a typo?

tools/cli/internal/openapi/file.go
```diff
@@ -31,20 +32,49 @@ const (
 	DotJSON = ".json"
 )
 
-func SaveSpec(path string, oas *Spec, format string, fs afero.Fs) error {
-	data, err := NewArrayBytesFromOAS(oas, path, format)
+// SaveToFile saves the content to a file in the specified format.
+// If format is empty, it saves the content in both JSON and YAML formats.
+// Note:
+// - SaveToFile expects the path to be a file path, not a directory path.
```
blva, line 38: [nit] if that's so, can we make sure to check if path is a directory and throw an error if it is?

tools/cli/internal/cli/changelog/create_test.go
```diff
@@ -0,0 +1,31 @@
...
+package changelog
+
+import (
+	"testing"
+
+	"github.com/mongodb/openapi/tools/cli/internal/cli/flag"
+	"github.com/mongodb/openapi/tools/cli/internal/test"
+)
+
+func TestCreateBuilder(t *testing.T) {
```
blva, line 24: [q] will we add more unit tests for create.go in another ticket?

tools/cli/internal/cli/changelog/create.go
```diff
@@ -81,6 +125,9 @@ func CreateBuilder() *cobra.Command {
 		Short:   "Generate the changelog for the OpenAPI spec.",
 		Args:    cobra.NoArgs,
 		RunE: func(_ *cobra.Command, _ []string) error {
+			if err := opts.validations(); err != nil {
```
blva, line 128: [q] why validations in RunE and not `PreRunE`?

tools/cli/internal/cli/changelog/create.go
```diff
@@ -45,19 +49,59 @@ func (o *Opts) Run() error {
...
+	if errSaveFile := openapi.SaveToFile(
+		o.newOutputFilePath(fmt.Sprintf("%s/%s", changelogAllFolderName, changelogAllFileName)), "", entries, o.fs); errSaveFile != nil {
+		return errSaveFile
+	}
+
+	if err := o.fs.MkdirAll(fmt.Sprintf("%s/%s", o.outputPath, versionChangelogFolderName), 0o755); err != nil {
+		return err
+	}
+
+	for _, entry := range versionedEntries {
```
blva, line 80: [q] not sure I remember why we need the versioned entries to bse saved, do we need them?

tools/cli/internal/openapi/file.go
andreaangiolillo: no, this comment is not needed. I am actually not sure what I meant here 🤦

tools/cli/internal/cli/changelog/create.go
andreaangiolillo: updated

tools/cli/internal/cli/changelog/create_test.go
andreaangiolillo: no, the command is only calling changelog functions and saving files. We already have unit tests for changelog and we are going also to add e2e tests so I think are okay. Let me know if you were thinking on specific unit tests scenarios that we could add for the create command and I will add them in a follow up PR

tools/cli/internal/cli/changelog/create.go
andreaangiolillo: my understanding is that the versioned entries (the `fromVersion_toVersion.json` files) are the ones shown to the user when comparing two versions

blva: hmmm, I don't remember, but I think that docs use changelog.json as well, but ok to keep them here until we find out

tools/cli/internal/cli/changelog/create_test.go
blva: yeah, I'd like to test the saving files logic for sure since we can find bugs also in the way we write files/folders etc - great if it can be a follow-up!

blva (approved):
LGTM, thanks for the changes!

PR #175, opened 2024-08-19
CLOUDP-263207 Use internal OAS instead of fetching from s3 bucket

Luke-Sanderson:
### Previous method
- Fetch a list of current versions and store in a file named versions.json
- Fetch the most recent version of the OAS from the s3 bucket.
- Compare OAS with cached sha to determine if workflow should continue
 
### Proposed method
- Only will start on Prod OAS Release PR merge so always will be a new version of the OAS
- Sort the files in the OpenAPI folder 
- Get the most recent OAS
- Extract the version from the file name and store in a file version.txt

### Advantages
- Removes 2 API calls per run
- Set to run only on merge of Prod OAS Release PR rather than Cron job

Tested on [fork](https://github.com/Luke-Sanderson/openapi/actions/runs/10455292810/job/28949901224), note it only failed on the upload step due to secrets not being set. Output on the fetch step shows correct version.

_Jira ticket:_ CLOUDP-263207

## Checklist


.github/workflows/release-postman.yml
```diff
@@ -1,39 +1,28 @@
 name: Postman Release
 on:
-  schedule:
-    - cron: '0 10 ? * *'
+  pull_request:
+    branches: 
+      - main
+    types:
+      - closed
```
andreaangiolillo, line 7: does this mean that this action will run after any new PRs that is merged into master?

.github/workflows/release-postman.yml
```diff
@@ -1,39 +1,28 @@
...
   workflow_dispatch:
 permissions:
   issues: write
 
 jobs:
   release-postman:
+    if: ${{ github.head_ref == 'api-bot-update' && github.event.pull_request.merged }}
```
andreaangiolillo, line 14: ah okay, here you are checking for PRs coming from the user used for the release

andreaangiolillo (approved):
LGTM

.github/workflows/release-postman.yml
Luke-Sanderson: Yeah it will run the workflow on every merge to master but the main job itself will only run on release PR merges. This was the only way I could find to do it on the [github docs](https://www.mongodb.com/developer/products/atlas/admin-api-postman/)

PR #181, opened 2024-08-20
CLOUDP-267019: [e2e test] Add an e2e scenario to test the changelog create [Part 2]

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-267019

Follow up from #179


tools/cli/test/data/changelog/same-api-version/output/2023-01-01_2023-02-01.json
```diff
@@ -0,0 +1,364 @@
+[
```
blva, line 1: is this the test output or the expected output?

blva: if test output, can we add to git ignore

andreaangiolillo: no, there are the files that I downloaded (manually) from evergreen and are compared to the ones generated by the e2e test

PR #184, opened 2024-08-20
feat: add `changelog metadata create` command

andreaangiolillo:
## Proposed changes
This PR creates the `changelog metadata create` to help with the generation of the metadata file during the release process

```bash
./bin/foascli changelog metadata create --help                                                                                         
Generate the changelog for the OpenAPI spec.

Usage:
  foascli changelog metadata create [flags]

Aliases:
  create, generate

Flags:
  -h, --help               help for create
  -o, --output string      File name or path where the command will store the output.
  -d, --run-date string    Date when the changelog was generated. (Format: YYYY-MM-DD)
      --sha string         SHA of the commit of the openapi specification used to generate the changelog.
  -v, --versions strings   List of versions to consider when generating the changelog. (Format: YYYY-MM-DD)
```

```bash
./bin/foascli changelog metadata create --sha e624d716e86f6910757b60cefdf3aa3181582d38 --versions=2024-01-01,2024-09-22             
{
  "runDate": "2024-08-20",
  "specRevision": "e624d716e86f6910757b60cefdf3aa3181582d38",
  "specRevisionShort": "e624d716e86",
  "versions": [
    "2024-01-01",
    "2024-09-22"
  ]
}
```

### DIFF
--- a/tools/cli/internal/cli/changelog/metadata/create_test.go
+++ b/tools/cli/internal/cli/changelog/metadata/create_test.go
@@ -26,7 +26,7 @@ import (
 func TestCreateBuild_Run(t *testing.T) {
 	fs := afero.NewMemMapFs()
 	opts := &Opts{
-		specRevision: "test",
+		specRevision: "11110c256dffdb163be71a3ca70854a57fad5f6f",
 		runDate:      "2024-01-01",
 		versions:     []string{"2024-01-01"},
 		fs:           fs,


tools/cli/internal/cli/changelog/metadata/metadata_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/mongodb/openapi/tools/cli/internal/test"
+)
+
+func TestBuilder(t *testing.T) {
+	test.CmdValidator(
+		t,
+		Builder(),
+		1,
```
jeroenvervaeke, line 27: [q] Does adding this kind of unit test add any value?

jeroenvervaeke (approved):

tools/cli/internal/cli/changelog/metadata/metadata_test.go
andreaangiolillo: for this command maybe not that much but when the command has flags it makes sure that all the expected flags are attached to the command

PR #238, opened 2024-10-07
CLOUDP-277321: reverted changelog changes

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-277321

This PR manually updates the prod changelog to remove the entries related to the invite endpoints.

FYI: This PR also fixes the changelog end-to-end (e2e) tests, which started failing because the newly generated changelog did not include the invite endpoints. As a result, the e2e tests were failing due to a mismatch between the generated changelog and the expected one.


blva (approved):
> FYI: This PR also fixes the changelog end-to-end (e2e) tests, which started failing because the newly generated changelog did not include the invite endpoints. As a result, the e2e tests were failing due to a mismatch between the generated changelog and the expected one.

e2e tests should not depend on our current changelog right? we can address in the TODO ticket?

tools/cli/test/data/changelog/same-api-version/base/openapi-2023-01-01.json
```diff
@@ -10495,7 +10495,7 @@
         } ],
         "summary" : "Return All Project Invitations",
         "tags" : [ "Projects" ],
-        "x-sunset" : "2024-10-04"
+        "x-sunset" : "2099-10-04"
```
andreaangiolillo, line 10498: > e2e tests should not depend on our current changelog right? we can address in the TODO ticket?

@blva 
The e2e test currently uses specs from MMS that have the sunset date set to 2024-10-04 (this is the change included in my PR). I’ll create a follow-up PR to update the e2e test specs to set the endpoint sunset date to 2099, so that we won’t encounter this error again.

blva: ah ok, it is a "frozen" spec but we reached the sunset date. fun! xD

PR #239, opened 2024-10-07
CLOUDP-277321: extend sunset date on e2e spec files

andreaangiolillo:
## Proposed changes

This PR fixes the changelog e2e tests to extend the sunset date of the specs used in the e2e tests to 2099.

Conversation: https://github.com/mongodb/openapi/pull/238#discussion_r1789986879

### DIFF
--- a/tools/cli/test/data/changelog/new-api-version/base/openapi-2023-01-01.json
+++ b/tools/cli/test/data/changelog/new-api-version/base/openapi-2023-01-01.json
@@ -946,7 +946,7 @@
         } ],
         "summary" : "Return One Identity Provider by ID",
         "tags" : [ "Federated Authentication" ],
-        "x-sunset" : "2025-01-01"
+        "x-sunset": "2099-01-01
       },
       "patch" : {
         "deprecated" : true,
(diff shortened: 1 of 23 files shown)


tools/cli/test/data/changelog/new-api-version/base/openapi-2023-01-01.json
```diff
@@ -946,7 +946,7 @@
         } ],
         "summary" : "Return One Identity Provider by ID",
         "tags" : [ "Federated Authentication" ],
-        "x-sunset" : "2025-01-01"
```
fmenezes, line 949: should we use the same dates from https://github.com/10gen/mms/pull/110515 ?

blva: maybe we can also add a ticket to add a filter functionality per checker to changelog, so that we could exclude/ignore the removed entries similar to what we use for the exemptions

andreaangiolillo: No, we shouldn’t do that. These files are used for e2e tests, and if we set that date, the tests will start failing again in 2026.

blva (approved):

PR #243, opened 2024-10-08
CLOUDP-271223: Skip conflict when x-xgen-soa-migration is set

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-271223

No doc diff + duplicated path
```
➜  cli git:(CLOUDP-271223) ✗ ./bin/foascli merge -b test/data/base_spec.json -e test/data/apiregistry_spec.json --output tes.json
2024/10/08 16:06:56 Detected x-xgen-soa-migration annotation in all operations for path:  /rest/v2/openapi/info
2024/10/08 16:06:56 Skipping conflict for path: /rest/v2/openapi/info
2024/10/08 16:06:56 No doc diff detected for path /rest/v2/openapi/info, merging the paths
2024/10/08 16:06:56 
We silently resolved the conflict with the schemas "OpenApiInfo" because the definition was identical.
2024/10/08 16:06:56 
We silently resolved the conflict with the schemas "info" because the definition was identical.
2024/10/08 16:06:56 
We silently resolved the conflict with the schemas "license" because the definition was identical.
2024/10/08 16:06:56 
We silently resolved the conflict with the schemas "ApiError" because the definition was identical.
2024/10/08 16:06:56 
File was saved in 'tes.json'.
```

Doc diff + duplicated path
```shell
➜  cli git:(CLOUDP-271223) ✗ ./bin/foascli merge -b test/data/base_spec.json -e test/data/apiregistry_spec.json --output tes.json
2024/10/08 17:11:06 Detected x-xgen-soa-migration annotation in all operations for path:  /rest/v2/openapi/info
2024/10/08 17:11:06 Skipping conflict for path: /rest/v2/openapi/info
Error: the path: "/rest/v2/openapi/info" is enabled for merge but it has a diff between the base and external spec. See the diff:
{
  "operations": {
    "modified": {
      "GET": {
        "operationID": {
          "from": "getOpenApiInaafo",
          "to": "getOpenApiInfo"
        }
      }
    }
  }
}
```
```shell
➜  cli git:(CLOUDP-271223) ✗ ./bin/foascli merge -b test/data/base_spec.json -e test/data/apiregistry_spec.json --output tes.json
2024/10/08 17:11:24 Detected x-xgen-soa-migration annotation in all operations for path:  /rest/v2/openapi/info
2024/10/08 17:11:24 Skipping conflict for path: /rest/v2/openapi/info
Error: the path: "/rest/v2/openapi/info" is enabled for merge but the flag to allow docs diff is not supported

```
## Checklist


tools/cli/internal/openapi/oasdiff.go
```diff
@@ -111,16 +113,18 @@ func (o OasDiff) mergePaths() error {
...
+	for path, externalPathData := range pathsToMerge.Map() {
+		// Tries to find if the path already exists or not
+		if originalPathData := basePaths.Value(path); originalPathData == nil {
+			basePaths.Set(path, removeExternalRefs(externalPathData))
 		} else {
-			return errors.PathConflictError{
-				Entry: k,
+			err := o.handlePathConflict(originalPathData, path)
+			if err != nil {
+				return err
```
andreaangiolillo, line 123:
```suggestion
			if err := o.handlePathConflict(originalPathData, path); err != nil{
				return err
```

tools/cli/internal/openapi/oasdiff.go
```diff
@@ -160,6 +164,69 @@ func removeExternalRefs(path *openapi3.PathItem) *openapi3.PathItem {
...
+	return errors.PathDocsDiffConflictError{
+		Entry: basePathName,
+		Diff:  d,
+	}
+}
+
+// shouldSkipConflict checks if the conflict should be skipped.
+// The method goes through each path operation and performs the following checks:
+// 1. Validates if both paths have same operations, if not, then it returns false.
+// 2. If both paths have the same operations, then it checks if there is an x-xgen-soa-migration annotation.
```
andreaangiolillo, line 207: Shouldn't the logic check for x-gen-soa-migration even if the operations are not identical? Scenario:
MMS has a resource with 4 operations and the standalone service has a resource with only one operation which has `x-xgen-soa-migration`. This scenario should be quite common as teams might migrate one operation at a time instead of all operations. Is this something that we should support?  
Let me know I am missing something and this scenario is covered

blva: Yes! :) You're right, I completely forgot that for GetOasPaths we give them the method + path, so it's posisble a path exists in both services. I'll just double-check with SRE, since our initial premisse was "no duplicated paths" (so maybe we'd ask teams to move everything all at once)

blva: It looks like if the answer is yes for my question, we'll also need to consider the path conflicts without x-xgen-soa-migration right? so I'll open a ticket for this functionality

blva: [CLOUDP-277871](https://jira.mongodb.org/browse/CLOUDP-277871)

tools/cli/internal/openapi/errors/merge_conflict_error.go
```diff
@@ -56,3 +61,26 @@ type TagConflictError struct {
...
+
+	return fmt.Sprintf("the path: %q is enabled for merge but it has a diff between the base and external spec. See the diff:\n%s", e.Entry, pathDiff)
+}
+
+type AllowDocsDiffNotSupportedError struct {
+	Entry string
+}
+
+func (e AllowDocsDiffNotSupportedError) Error() string {
+	return fmt.Sprintf("the path: %q is enabled for merge but the flag to allow docs diff is not supported", e.Entry)
```
andreaangiolillo, line 85: is there a flag to enable this support? if yes, could you add it to the error message?

tools/cli/internal/openapi/errors/merge_conflict_error.go
```diff
@@ -56,3 +61,26 @@ type TagConflictError struct {
...
+}
+
+func (e PathDocsDiffConflictError) Error() string {
+	var pathDiff []byte
+	_, ok := e.Diff.PathsDiff.Modified[e.Entry]
+	if ok {
+		pathDiff, _ = json.MarshalIndent(e.Diff.PathsDiff.Modified[e.Entry], "", "  ")
+	}
+
+	return fmt.Sprintf("the path: %q is enabled for merge but it has a diff between the base and external spec. See the diff:\n%s", e.Entry, pathDiff)
```
andreaangiolillo, line 77: could you also provide the two files involved in this mismatch? We have a ticket to improve the error message of a conflict to list also the files that are involved to help investigate the issue.

tools/cli/internal/openapi/oasdiff.go
```diff
@@ -22,16 +22,18 @@ import (
...
-	base     *load.SpecInfo
-	external *load.SpecInfo
-	config   *diff.Config
-	specDiff *diff.Diff
-	parser   Parser
+	base      *load.SpecInfo
+	external  *load.SpecInfo
+	config    *diff.Config
+	specDiff  *diff.Diff
+	noExtDiff NoExtensionDiff
```
andreaangiolillo, line 35: The name `NoExtensionDiff` gives fom granted a lot of context on how we use extensions in the spec. Main question from a person without context is what is the difference between `specDiff *diff.Diff` and `noExtDiff NoExtensionDiff`? 

Question: can we have an interface that includes both `specDiff *diff.Diff` and `noExtDiff NoExtensionDiff` as they are both checking differences between specifications

tools/cli/internal/openapi/oasdiff.go
```diff
@@ -160,6 +163,69 @@ func removeExternalRefs(path *openapi3.PathItem) *openapi3.PathItem {
...
+	if pathsAreIdentical, err = o.arePathsIdenticalWithExcludeExtensions(basePathName); err != nil {
+		return err
+	}
+
+	log.Printf("Skipping conflict for path: %s, pathsAreIdentical: %v", basePathName, pathsAreIdentical)
+	if pathsAreIdentical {
+		return nil
+	}
+
+	// allowDocsDiff = true not supported
```
andreaangiolillo, line 185: Could you update this comment to add more context: what is `allowDocsDiff`?, why is not supported?  what is the implication in this piece of code?

tools/cli/internal/openapi/oasdiff.go
```diff
@@ -417,6 +483,27 @@ func (o OasDiff) areSchemaIdentical(name string) bool {
 	return !ok
 }
 
+// arePathsIdenticalWithExcludeExtensions checks if the paths are identical with the extensions excluded
```
andreaangiolillo, line 486: could you provide an example of the extension in the comment?

tools/cli/internal/openapi/oasdiff.go
```diff
@@ -417,6 +483,27 @@ func (o OasDiff) areSchemaIdentical(name string) bool {
...
+func (o OasDiff) arePathsIdenticalWithExcludeExtensions(name string) (bool, error) {
```
andreaangiolillo, line 487: the code seems to be checking if the path are identical with and without extensions 🤔  is this correct? if yes , we should rename the function to reflect this logic

tools/cli/internal/openapi/oasdiff_test.go
```diff
@@ -1142,3 +1145,198 @@ func TestUpdateExternalRefReqBody(t *testing.T) {
...
+			},
+			expectedError: nil,
+		},
+		{
+			name: "Conflict with AllowDocsDiff",
+			basePath: &openapi3.PathItem{
+				Get: &openapi3.Operation{
+					Extensions: map[string]interface{}{
+						"x-xgen-soa-migration": map[string]interface{}{
+							"allowDocsDiff": "true",
```
andreaangiolillo, line 1182: if we have  `AllowDocsDiffNotSupportedError`, why allowDocsDiff can be set to true?

tools/cli/internal/openapi/paths.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+//go:generate mockgen -destination=../openapi/mock_paths.go -package=openapi github.com/mongodb/openapi/tools/cli/internal/openapi NoExtensionDiff
+
+const (
+	xgenSoaMigration = "x-xgen-soa-migration"
+	allowDocsDiff    = "allowDocsDiff"
+)
+
+// allOperationsHaveExtension checks if all the operations in the base pat have the given extension name.
```
andreaangiolillo, line 32:
```suggestion
// allOperationsHaveExtension checks if all the operations in the base path have the given extension name.
```

tools/cli/internal/openapi/paths.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+
+	if basePath.Delete != nil {
+		if result := getOperationExtensionWithName(basePath.Delete, extensionName); result == nil {
+			return false
+		}
+	}
+
+	log.Println("Detected x-xgen-soa-migration annotation in all operations for path: ", basePathName)
```
andreaangiolillo, line 68: my understanding is that this function can be used for all type of extension. Is this correct? if yes, you should update this log message to use the extension name in the input.

tools/cli/internal/openapi/paths.go
```diff
@@ -0,0 +1,153 @@
...
+func getOperationExtensionWithName(operation *openapi3.Operation, extensionName string) interface{} {
+	if operation.Extensions == nil || operation.Extensions[extensionName] == nil {
+		log.Printf("Operation %s does not have extension %q", operation.OperationID, extensionName)
+		return nil
+	}
+
+	return operation.Extensions[extensionName]
+}
+
+func allOperationsAllowDocsDiff(basePath *openapi3.PathItem) bool {
```
andreaangiolillo, line 81: all of these functions seem to share the logic to check for something in each resource operation. Could you just have a function that runs multiple checks for each operation?

tools/cli/internal/openapi/paths.go
blva: thanks! updated

tools/cli/internal/openapi/errors/merge_conflict_error.go
blva: no, this will be enabled in a follow-up PR, original idea was to not allow docs diff, but it came up as a possible functionality so it's low priority. https://jira.mongodb.org/browse/CLOUDP-277977

tools/cli/internal/openapi/oasdiff.go
blva: starting with decoupling specDiff since we already have an oasdiff result struct in https://github.com/mongodb/openapi/pull/250

tools/cli/internal/openapi/oasdiff_test.go
blva: I think answer is the same as it's a separate ticket to support it (cause initially we were not going to support)

tools/cli/internal/openapi/oasdiff.go
blva: this runs oasdiff with an "ignore extensions" config, I've updated the comment

tools/cli/internal/openapi/paths.go
blva: can cover as a follow up as we have many other places in the code where we perform checks for each operation

tools/cli/internal/openapi/oasdiff.go
blva: updated!

andreaangiolillo (approved):
LGTM, thanks for all the clarifications!

PR #260, opened 2024-10-17
CLOUDP-278929: Send a remind if an API version is approching the release date

andreaangiolillo:
## Proposed changes
This PR introduces a new GitHub Action that runs daily to verify that the API versions listed in the versions.json file in the development environment are not nearing their release dates (less than three weeks away). If there are any upcoming API versions scheduled for release, the action performs the following steps:
- Check if there is already a Jira ticket for the API version
   - If there is already a ticket, the action will stop here
- Create a jira ticket
- Send a slack notification to the team channel.

_Jira ticket:_ CLOUDP-278929

## Note
The only way to test the action is by running it after the merge, so I will open follow-up PRs to fix possible bugs.


.github/scripts/upcoming_api_releases.sh
```diff
@@ -0,0 +1,48 @@
...
+
+    # Check if the date is within 3 weeks (21 days)
+    if [[ "${diff_in_days}" -ge 0 && "${diff_in_days}" -le 21 ]]; then
+        # Add the date to the list if within 3 weeks
+        dates_within_3_weeks+=("${date}")
+    fi
+done
+
+if [[ ${#dates_within_3_weeks[@]} -gt 0 ]]; then
+    echo "API Versions that will be release in the next 3 weeks: ${dates_within_3_weeks[*]}"
```
tibulca, line 44: this will get quite noisy (every day, for 3 weeks in a row, we'll get this notification and tag @apix-2-on-call).
Is it too complicated to have this trigger conditions?
- isMonday and an API version is going to be released withing the next 3 weeks
- isWeekday and an API version is going to be released withing the 6 days (this is the release week)

andreaangiolillo: > isWeekday and an API version is going to be released withing the 6 days (this is the release week)

Isn't this still too noisy during the release week? I think that having the message only on Monday should be enough. WDYT?

andreaangiolillo: updated

.github/scripts/upcoming_api_releases.sh
```diff
@@ -0,0 +1,48 @@
...
+set -eou pipefail
+
+
+URL="https://raw.githubusercontent.com/mongodb/openapi/dev/openapi/v2/versions.json"
+
+# Fetch the version.json file
+response=$(curl -s "${URL}")
+
+# Parse the dates from the JSON response using jq
+dates=$(echo "${response}" | jq -r '.[]')
```
tibulca, line 11:
```suggestion
version_dates=$(echo "${response}" | jq -r '.[]')
```

tibulca: same for the dependent vars

andreaangiolillo: updated

.github/scripts/upcoming_api_releases.sh
```diff
@@ -0,0 +1,59 @@
...
+#    a. Converts the date to seconds since epoch.
+#    b. Calculates the difference in days between the date and the current date.
+#    c. Checks if the date is within 3 weeks (21 days) and adds it to the list if it is.
+# 5. Outputs the API versions that will be released in the next 3 weeks to the GitHub Actions output variable if any are found.
+
+
+URL="https://raw.githubusercontent.com/mongodb/openapi/dev/openapi/v2/versions.json"
+
+# Fetch the version.json file
+response=$(curl -s "${URL}")
```
blva, line 19: wondering what will be our plan long-term with .sh logic in our platform since in the past it has proven itself difficult to maintain, perharps we can add unit tests here like it was explored with the postman project? OK to make it in a follow-up PR

.github/scripts/upcoming_api_releases.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+set -eou pipefail
+
+# This script checks for upcoming API version releases within the next 3 weeks.
+# It performs the following steps:
+# 1. Fetches and oarse the `versions.json` file using in the `mongodb/openapi` repository deom the `dev` branch
```
blva, line 6:
```suggestion
# 1. Fetches and parses the `versions.json` file using in the `mongodb/openapi` repository on the `dev` branch
```

blva (approved):

.github/scripts/upcoming_api_releases.sh
andreaangiolillo: I will create a ticket to investigate how much time we might need to migrate these scripts to golang

PR #274, opened 2024-11-12
CLOUDP-284014: Check soa migration in external path

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-284014

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/cli/internal/openapi/oasdiff.go
+++ b/tools/cli/internal/openapi/oasdiff.go
@@ -76,7 +76,7 @@ func (o OasDiff) mergePaths() error {
 		if originalPathData := basePaths.Value(path); originalPathData == nil {
 			basePaths.Set(path, removeExternalRefs(externalPathData))
 		} else {
-			if err := o.handlePathConflict(originalPathData, path); err != nil {
+			if err := o.handlePathConflict(externalPathData, path); err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


tools/cli/internal/openapi/oasdiff.go
```diff
@@ -76,7 +76,7 @@ func (o OasDiff) mergePaths() error {
 		if originalPathData := basePaths.Value(path); originalPathData == nil {
 			basePaths.Set(path, removeExternalRefs(externalPathData))
 		} else {
-			if err := o.handlePathConflict(originalPathData, path); err != nil {
+			if err := o.handlePathConflict(externalPathData, path); err != nil {
```
blva, line 79: main change: we were getting the originalPathData (baseData, which is mms, but we expect annotations to be in SOA)

andreaangiolillo: What does it mean "_expect annotations to be in SOA_"? Do you mean the file path of the spec was MMS but we need the service spec instead?

blva: expect annotations to be in Standalone service, correct. We were adding the annotation to MMS, but it should live in the standalone service once the copy endpoint is created

andreaangiolillo (approved):
LGTM

PR #281, opened 2024-11-12
CLOUDP-279336: Avoid that squashing overrides hideFromChangelog

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-279336

Similar to this: https://github.com/10gen/mms/blob/2bdd398c7ede9d7d1e58519c6931fb1dbae13a36/scripts/api_versioning/changelog/generate_changelog.py#L197-L204

As part of investigating exemptions, I also noticed we weren't accounting for hiddenFromChangelog values while squashing, which could end up overriding the hide value. I'm following a similar logic to the one in current changelog above to fix it and updated all tests. 

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/cli/internal/changelog/outputfilter/squash.go
+++ b/tools/cli/internal/changelog/outputfilter/squash.go
@@ -135,11 +135,16 @@ func squashEntries(entries []*OasDiffEntry) ([]*OasDiffEntry, error) {
 	entriesMap := newEntriesMapPerIDAndOperationID(entries)
 	squashHandlers := newSquashHandlers()
 	squashedEntries := []*OasDiffEntry{}
+	hidddenSquashedEntries := []*OasDiffEntry{}
 
 	for _, entry := range entries {
 		// if no squash handlers implemented for entry's code,
 		// just append the entry to the result
 		if _, ok := findHandler(entry.ID); !ok {
+			if entry.HideFromChangelog {
+				hidddenSquashedEntries = append(hidddenSquashedEntries, entry)
+				continue
+			}
 			squashedEntries = append(squashedEntries, entry)
 			continue
 		}
@@ -156,6 +161,7 @@ func squashEntries(entries []*OasDiffEntry) ([]*OasDiffEntry, error) {
 			return nil, err
 		}
 
+		squashedEntries = append(squashedEntries, hidddenSquashedEntries...)
 		squashedEntries = append(squashedEntries, sortEntriesByDescription(entries)...)
 	}
 
--- a/tools/cli/internal/changelog/outputfilter/squash.go
+++ b/tools/cli/internal/changelog/outputfilter/squash.go
@@ -111,7 +111,8 @@ func newSquashHandlers() []handler {
 	}
 }
 
-// EntryMappings groups entries by ID and then by OperationID and returns a Map[changeCode, Map[operationId, List[oasdiffEntry]]]
+// EntryMappings groups entries by ID and then by OperationID and returns two maps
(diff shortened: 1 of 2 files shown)


tools/cli/internal/changelog/outputfilter/squash.go
```diff
@@ -112,27 +112,40 @@ func newSquashHandlers() []handler {
 }
 
 // EntryMappings groups entries by ID and then by OperationID and returns a Map[changeCode, Map[operationId, List[oasdiffEntry]]]
-func newEntriesMapPerIDAndOperationID(entries []*OasDiffEntry) map[string]map[string][]*OasDiffEntry {
-	result := make(map[string]map[string][]*OasDiffEntry)
+func newEntriesMapPerIDAndOperationID(entries []*OasDiffEntry) (result, hiddenResult map[string]map[string][]*OasDiffEntry) {
```
andreaangiolillo, line 115: could you update the function comment to include the new returned map?

tools/cli/internal/changelog/outputfilter/squash.go
```diff
@@ -112,27 +112,40 @@ func newSquashHandlers() []handler {
...
+		hidden := entry.HideFromChangelog
 
 		// Ensure the code map exists
 		if _, exists := result[code]; !exists {
 			result[code] = make(map[string][]*OasDiffEntry)
 		}
 
+		if _, exists := hiddenResult[code]; !exists {
+			hiddenResult[code] = make(map[string][]*OasDiffEntry)
+		}
```
andreaangiolillo, line 131: Shouldn't you create a new entry in the map only if there is at least one entry hidden? if yes., you could come this inside the hidden if

tools/cli/internal/changelog/outputfilter/squash.go
```diff
@@ -145,6 +158,24 @@ func squashEntries(entries []*OasDiffEntry) ([]*OasDiffEntry, error) {
...
+		return nil, err
+	}
+
+	squashedEntriesHidden, err := appplySquashHandlerToMap(squashHandlers, hiddenEntriesByIDandOperationID)
+	if err != nil {
+		return nil, err
+	}
+
+	squashedEntries = append(squashedEntries, squashedEntriesNotHidden...)
+	squashedEntries = append(squashedEntries, squashedEntriesHidden...)
```
andreaangiolillo, line 172: since we append the hidden entries in the same array, how do we then know that the hidden should not be in the changelog.json but only in the chanelog-all.json?

andreaangiolillo (commented):
Only one open question, the other comments are not blocking

tools/cli/internal/changelog/outputfilter/squash.go
blva: we could but the way its being done for the visible ones is like this, both end up adding the code into the map before, even if there are no entries. in that case I can do that for both, I just wasn't sure if this statement existed to cover some corner case I'm not aware of

tools/cli/internal/changelog/outputfilter/squash.go
blva: my understanding is that the differentiation between changelog.json and changelog-all.json happens later, at the `changelog.go` level in `NewNotHiddenEntries`

andreaangiolillo: yes, good point. I noticed that the OasDiffEntry has a field HideFromChangelog that will be used for this purpose. Thanks !

andreaangiolillo (approved):
LGTM

tools/cli/internal/changelog/outputfilter/squash.go
blva: I'm making the change and not finding any failures so lets try that!

PR #284, opened 2024-11-13
CLOUDP-284437: Add Spectral rule to ensure operationId uses camelcase

andreaangiolillo:
## Proposed changes

_Jira ticket:_ CLOUDP-284437

### DIFF
--- a/tools/spectral/.spectral.yaml
+++ b/tools/spectral/.spectral.yaml
@@ -124,7 +124,7 @@ rules:
 
   xgen-docs-operationId-camel-case:
     description: operationId should be camelCased.
-    given: "$.operationId"
+    given: "$.paths[*]..operationId"
     severity: error
     then:
       function: casing
--- a/tools/spectral/.spectral.yaml
+++ b/tools/spectral/.spectral.yaml
@@ -124,7 +124,7 @@ rules:
 
   xgen-docs-operationId-camel-case:
     description: operationId should be camelCased.
-    given: "operationId"
+    given: "$.operationId"
     severity: error
     then:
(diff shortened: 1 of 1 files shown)


tools/spectral/.spectral.yaml
```diff
@@ -122,6 +122,15 @@ rules:
       functionOptions:
         match: "^[A-Za-z0-9-._~:/?#\\[\\]@!\\$&'()*+,;=]*$"
 
+  xgen-docs-operationId-camel-case:
+    description: operationId should be camelCased.
+    given: "$.paths[*]..operationId"
+    severity: error
+    then:
```
gssbzn, line 129: can you add a message like 
```suggestion
    message: "Operation ID should be camelCase: {{property}}."
    then:
```
so we can see what the current value is?

andreaangiolillo: good point! I will update it

andreaangiolillo: I updated to use `{{value}}` which gives the message "40221:26    error  xgen-docs-operationId-camel-case  Operation ID should be camelCase: grantMongoDBEmployeeAccess."

Using `{{property}}`` does not give a useful message  "40221:26    error  xgen-docs-operationId-camel-case  Operation ID should be camelCase: operationId."

gssbzn: perfect I wasn't sure what the right interpolation was, thanks for fixing it

lovisaberggren (approved):

blva (approved):

PR #308, opened 2024-12-12
CLOUDP-276602: FOASCLI: Improve the error message in the event of conflicts

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-276602](https://jira.mongodb.org/browse/CLOUDP-276602)

This PR proposes to update the conflict error messages to include the Base and external spec locations to help debug where the issue is.

### DIFF
--- a/tools/cli/test/e2e/cli/merge_test.go
+++ b/tools/cli/test/e2e/cli/merge_test.go
@@ -103,7 +103,7 @@ func TestMerge(t *testing.T) {
 		require.Error(t, err, stringResponse)
 		assert.Contains(t, stringResponse, fmt.Sprintf("Error: there was a conflict with the Tag \"Events\""+
 			" with the description: \"Returns information about the MongoDB Atlas Specification.\"."+
-			" Base Spec: %q, External Spec: %q", base, apiRegistrySpec))
+			" Base Spec: %q, External Spec: %q", base, authnSpec))
 	})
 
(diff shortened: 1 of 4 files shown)


tools/cli/test/e2e/cli/merge_test.go
```diff
@@ -122,6 +125,9 @@ func TestMerge(t *testing.T) {
 		resp, err := cmd.CombinedOutput()
 		stringResponse := string(resp)
 		require.Error(t, err, stringResponse)
-		assert.Contains(t, stringResponse, "Error: there was a conflict on a Schema component: \"ApiError\"")
+		assert.Contains(t, stringResponse,
```
blva, line 128: can we display the diff between the two schemas?

blva (approved):
lgtm, would be good if we could display the diff

tools/cli/test/e2e/cli/merge_test.go
andreaangiolillo: I think we'd risk printing an enormous string if we do so since the struct used to get the diff is SchemaDiff which has +20 fields + sub-fields

blva: we could output in a file maybe then and leave it in the files if the person wants to get details, but no strong opinion feel free to ignore

PR #313, opened 2024-12-17
CLOUDP-287245: IPA-109: Validate custom method must be GET or POST

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ [CLOUDP-287245](https://jira.mongodb.org/browse/CLOUDP-287245)

Implements a rule that the HTTP method for custom methods must be either GET or POST (https://docs.devprod.prod.corp.mongodb.com/ipa/109).

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/spectral/ipa/rulesets/IPA-109.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-109.yaml
@@ -13,7 +13,3 @@ rules:
     then:
       field: '@key'
       function: 'eachCustomMethodMustBeGetOrPost'
-      functionOptions:
-        requiredMethods:
-          - get
-          - post
--- a/tools/spectral/ipa/rulesets/IPA-109.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-109.yaml
@@ -1,0 +1,19 @@
+# IPA-109: Custom Methods
+# http://go/ipa/109
+
+functions:
+  - eachCustomMethodMustBeGetOrPost
+
+rules:
+  xgen-IPA-109-custom-method-must-be-GET-or-POST:
+    description: "The HTTP method for custom methods must be GET or POST. http://go/ipa/109"
+    message: "{{error}} http://go/ipa/109"
+    severity: warn
+    given: "$.paths"
+    then:
+      field: "@key"
+      function: "eachCustomMethodMustBeGetOrPost"
+      functionOptions:
+        requiredMethods:
+          - get
+          - post
--- a/tools/spectral/ipa/rulesets/IPA-109.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-109.yaml
@@ -6,13 +6,13 @@ functions:
 
 rules:
   xgen-IPA-109-custom-method-must-be-GET-or-POST:
-    description: "The HTTP method for custom methods must be GET or POST. http://go/ipa/109"
-    message: "{{error}} http://go/ipa/109"
+    description: 'The HTTP method for custom methods must be GET or POST. http://go/ipa/109'
+    message: '{{error}} http://go/ipa/109'
     severity: warn
-    given: "$.paths"
+    given: '$.paths'
     then:
-      field: "@key"
-      function: "eachCustomMethodMustBeGetOrPost"
+      field: '@key'
+      function: 'eachCustomMethodMustBeGetOrPost'
       functionOptions:
         requiredMethods:
           - get
--- a/tools/spectral/ipa/rulesets/IPA-109.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-109.yaml
@@ -9,6 +9,6 @@ rules:
     description: 'The HTTP method for custom methods must be GET or POST. http://go/ipa/109'
     message: '{{error}} http://go/ipa/109'
     severity: warn
-    given: '$.paths'
+    given: '$.paths[*]'
     then:
       function: 'eachCustomMethodMustBeGetOrPost'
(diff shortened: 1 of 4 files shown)


tools/spectral/ipa/rulesets/IPA-109.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+rules:
+  xgen-IPA-109-custom-method-must-be-GET-or-POST:
+    description: 'The HTTP method for custom methods must be GET or POST. http://go/ipa/109'
+    message: '{{error}} http://go/ipa/109'
+    severity: warn
+    given: '$.paths'
+    then:
+      field: '@key'
+      function: 'eachCustomMethodMustBeGetOrPost'
```
lovisaberggren, line 15:
```suggestion
    given: '$.paths'
    then:
      function: 'eachCustomMethodMustBeGetOrPost'
```
Since we are evaluating the path items in this case, you can pass them directly here, and then in the function you don't need to retrieve them from the documentInventory

tools/spectral/ipa/rulesets/functions/eachCustomMethodMustBeGetOrPost.js
```diff
@@ -0,0 +1,33 @@
...
+
+const ERROR_MESSAGE = 'The HTTP method for custom methods must be GET or POST.';
+const VALID_METHODS = ['get', 'post'];
+
+export default (paths) => {
+  // Collect all errors
+  const errors = [];
+
+  // Iterate through each path key and its corresponding path item
+  for (const [pathKey, pathItem] of Object.entries(paths)) {
```
lovisaberggren, line 11: Ah didn't realize that `paths` is not an array, I thought that the change would only make the input contain a single path and it's pathitems, not the whole `paths` object. Sorry about that! This makes the function a bit harder to read IMO, feel free to revert back to the previous approach if you want. WDYT?

lovisaberggren: Or perhaps try `$.paths[*]`

yelizhenden-mdb: It worked with `$.paths[*]` and using `path` from the `context`. Could you take another look and let me know how it looks?

yelizhenden-mdb: `$.paths[*]` gives you the object for each path, but it doesn't include the path keys themselves, so it's not quite enough on its own

lovisaberggren: Thanks! Looks good! Just one question on the logic

tools/spectral/ipa/rulesets/functions/eachCustomMethodMustBeGetOrPost.js
```diff
@@ -0,0 +1,26 @@
...
+  let pathKey = path[1];
+
+  if (!isCustomMethod(pathKey)) return;
+
+  const httpMethods = Object.keys(input);
+
+  // Check for invalid methods
+  if (httpMethods.some((method) => !VALID_METHODS.includes(method))) {
+    return ERROR_RESULT;
+  }
```
lovisaberggren, line 18: Will this return error if the path has an exception extension? 

Suggestion: use `$.path[get,put,post,delete,options,head,patch,trace]` to ignore any other objects than the methods, kinda like we do here: https://github.com/mongodb/openapi/blob/98f5b237e447fc07b191833195a32fc310a4daa4/tools/spectral/.spectral.yaml#L9

yelizhenden-mdb: Yes, it would return error with an exception extension. Good catch!

yelizhenden-mdb: `"#PathItem[get,put,post,delete,options,head,patch,trace]" ` gives the object inside HTTP method. I will exclude `x-xgen-IPA-exception`

tools/spectral/ipa/rulesets/functions/eachCustomMethodMustBeGetOrPost.js
```diff
@@ -0,0 +1,29 @@
...
+
+export default (input, opts, { path }) => {
+  // Extract the path key (e.g., '/a/{exampleId}:method') from the JSONPath.
+  let pathKey = path[1];
+
+  if (!isCustomMethod(pathKey)) return;
+
+  //Exclude exception extension key
+  let keys = Object.keys(input);
+  const httpMethods = keys.filter((key) => key !== EXCEPTION_EXTENSION_KEY);
```
yelizhenden-mdb, line 16: @lovisaberggren I excluded the exception extension key here, but in the opposite I can only check the list of predefined HTTP methods as well

lovisaberggren: Thanks! Yeah perhaps the opposite is more stable, as there may be other extensions for example added to paths, and HTTP methods probably won't change a lot 😄

tools/spectral/ipa/rulesets/functions/eachCustomMethodMustBeGetOrPost.js
```diff
@@ -0,0 +1,29 @@
+import { isCustomMethod } from './utils/resourceEvaluation.js';
+
+const ERROR_MESSAGE = 'The HTTP method for custom methods must be GET or POST.';
+const ERROR_RESULT = [{ message: ERROR_MESSAGE }];
+const VALID_METHODS = ['get', 'post'];
+const HTTP_METHODS = ['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace'];
```
lovisaberggren, line 6: [Just a note for future rules]: I expect we will have other similar rules coming up, at that point we can consider refactoring out a common method to check for valid/invalid methods to the utils package

lovisaberggren (approved):
Thanks! 🚀

PR #321, opened 2024-12-18
CLOUDP-288860: Support IPA exception and validate exception extensions

lovisaberggren:
## Proposed changes

Adds ability to ignore API components with IPA exception extensions to existing IPA rules:

- `xgen-IPA-104-resource-has-GET`
- `xgen-IPA-102-path-alternate-resource-name-path-param`

Also added a IPA rule validation `xgen-IPA-005-exception-extension-format` to check the format of any `x-xgen-IPA-exception` extensions. This rule does not accept exceptions, and should not.

_Jira ticket:_ [CLOUDP-288860](https://jira.mongodb.org/browse/CLOUDP-288860)

### Testing

- Added tests for existing rules to validate that the component is ignored when an `x-xgen-IPA-exception` extension is present with the corresponding rule name
- Added tests for the new `xgen-IPA-005-exception-extension-format` rule to validate that it catches invalid exception extensions
- Unit test for `hasException` helper function

### DIFF
--- a/tools/spectral/ipa/rulesets/functions/exceptionExtensionFormat.js
+++ b/tools/spectral/ipa/rulesets/functions/exceptionExtensionFormat.js
@@ -1,0 +1,31 @@
+const ERROR_MESSAGE = 'IPA exceptions must have a valid rule name and a reason.';
+const RULE_NAME_PREFIX = 'xgen-IPA-';
+const REASON_KEY = 'reason';
+
+// Note: This rule does not allow exceptions
+export default (input, _, { path }) => {
+  const exemptedRules = Object.keys(input);
+  const errors = [];
+
+  exemptedRules.forEach((ruleName) => {
+    const exception = input[ruleName];
+    if (!isValidException(ruleName, exception)) {
+      errors.push({
+        path: path.concat([ruleName]),
+        message: ERROR_MESSAGE,
+      });
+    }
+  });
+
+  return errors;
+};
+
+function isValidException(ruleName, exception) {
+  const exceptionObjectKeys = Object.keys(exception);
+  return (
+    ruleName.startsWith(RULE_NAME_PREFIX) &&
+    exceptionObjectKeys.length === 1 &&
+    exceptionObjectKeys.includes(REASON_KEY) &&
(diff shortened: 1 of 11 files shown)


tools/spectral/ipa/rulesets/functions/exceptionExtensionFormat.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+  });
+
+  return errors;
+};
+
+function isValidException(ruleName, exception) {
+  const exceptionObjectKeys = Object.keys(exception);
+  return (
+    ruleName.startsWith(RULE_NAME_PREFIX) &&
```
yelizhenden-mdb, line 26: Q: Should we check if the `ruleName` is among the rules defined?

tools/spectral/ipa/rulesets/functions/utils/exceptions.js
```diff
@@ -0,0 +1,15 @@
+const EXCEPTION_EXTENSION = 'x-xgen-IPA-exception';
+
+/**
+ * Checks if the object has an exception extension "x-xgen-IPA-exception"
+ *
+ * @param object the object to evaluate
+ * @param ruleName the name of the exempted rule
+ * @returns {boolean} true if the object has an exception named ruleName, otherwise false
+ */
+export function hasException(object, ruleName) {
```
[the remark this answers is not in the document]
lovisaberggren: Yeah I decided not to check the format here since we have a rule that covers it already, so here we assume the extension has keys with the rule names

tools/spectral/ipa/rulesets/functions/exceptionExtensionFormat.js
yelizhenden-mdb: Yes, I think you would need another function which loads `ipa-spectral.yaml` to a Spectral object and fetch the rule names from there

yelizhenden-mdb: I think it is fine by now, it can be an improvement for later

yelizhenden-mdb (approved):
LGTM 🚀

PR #350, opened 2025-01-10
CLOUDP-261180: Add filter to remove unused views

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [CLOUDP-261180](https://jira.mongodb.org/browse/CLOUDP-261180)

This PR adds a new filter to remove the unused schemas in the components section of the openapi spec.


wtrocki:
@andreaangiolillo Is this still in your plans? That will significantly reduce openapi size. We could then remove several filters in SDK-go repository that try to archieve the same.

tools/cli/test/data/split/dev/openapi-v2-2023-01-01.json
```diff
@@ -1,15 +1,4 @@
 {
-  "info": {
```
andreaangiolillo, line 2: I needed to update the specifications used for the e2e split command. The issue was that, due to the filters applied, the generated specs were missing unused schemas. This caused the test to fail because the generated specs were not identical to the expected ones.

fmenezes:
[q] how about the main spec with all versions?

tools/cli/internal/openapi/filter/schemas.go
```diff
@@ -0,0 +1,65 @@
...
+				schemasToDelete = append(schemasToDelete, k)
+			}
+		}
+
+		if len(schemasToDelete) == 0 {
+			return nil
+		}
+
+		for _, schemaToDelete := range schemasToDelete {
+			log.Printf("Deleting unused schema: '%s'", schemaToDelete)
```
gssbzn, line 59:
```suggestion
			log.Printf("Deleting unused schema: %q", schemaToDelete)
```

andreaangiolillo: updated

andreaangiolillo:
> [q] how about the main spec with all versions?

As part of the release process, we generate the `v2.json` specification containing all API versions. Then, we use the `foascli filter` command to clean up the spec by removing extensions and, after this PR, unused views as well. The `foascli filter` command relies on the filter list updated in this PR. Once we release the updated version of `foascli`, it will automatically start using the new filter configuration. @fmenezes

gssbzn (approved):

PR #352, opened 2025-01-13
CLOUDP-290417: [Product Metrics/Observability] Implement Collector class

yelizhenden-mdb:
## Proposed changes

Implements a Collector class which collects all the exceptions, violations and adoptions of IPA rules as an object of the component itself and the name of the related IPA rule.

It outputs all of the results to a JSON file

_Jira ticket:_ [CLOUDP-290417](https://jira.mongodb.org/browse/CLOUDP-290417)

## Testing
When you run `spectral lint` command with the rulesets in `ipa-spectral.yaml` file, it will output a `combined.log` file which will collect all the exceptions, violations and adoptions.

Example file [ipa-collector-results-combined.log](https://github.com/user-attachments/files/18411345/ipa-collector-results-combined.log), after the spectral linting on `openapi/v2.json` file
You can navigate to the lists of `exceptions`, `violations` and `adoptions`

## Checklist

### Changes to Spectral

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,4 @@
 *.vscode
 
 *.out
-/tools/spectral/ipa/ipa-collector-results-combined.log
+**/*ipa-collector-results-combined.log
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@
 *.vscode
 
 *.out
+/tools/spectral/ipa/ipa-collector-results-combined.log
--- a/tools/spectral/ipa/metrics/collector.js
+++ b/tools/spectral/ipa/metrics/collector.js
@@ -27,7 +27,7 @@ class Collector {
       [EntryType.EXCEPTION]: [],
     };
 
-    this.fileName = "combined.log"
+    this.fileName = 'combined.log';
 
     process.on('exit', () => this.flushToFile());
     process.on('SIGINT', () => {
@@ -37,9 +37,9 @@ class Collector {
   }
 
   add(componentId, ruleName, type) {
-    if(componentId && ruleName && type) {
+    if (componentId && ruleName && type) {
       componentId = componentId.join('.');
-      const data = {componentId, ruleName};
+      const data = { componentId, ruleName };
       this.entries[type].push(data);
     }
   }
--- a/tools/spectral/ipa/metrics/collector.js
+++ b/tools/spectral/ipa/metrics/collector.js
@@ -27,7 +27,7 @@ class Collector {
       [EntryType.EXCEPTION]: [],
     };
 
-    this.fileName = 'combined.log';
+    this.fileName = 'ipa-collector-results-combined.log';
 
     process.on('exit', () => this.flushToFile());
     process.on('SIGINT', () => {
@@ -36,11 +36,20 @@ class Collector {
     });
   }
 
-  add(componentId, ruleName, type) {
+  add(type, componentId, ruleName, exceptionReason = null ) {
     if (componentId && ruleName && type) {
+      if (!Object.values(EntryType).includes(type)) {
+        throw new Error(`Invalid entry type: ${type}`);
+      }
+
       componentId = componentId.join('.');
-      const data = { componentId, ruleName };
-      this.entries[type].push(data);
+      const entry = {componentId, ruleName};
+
+      if(type === EntryType.EXCEPTION && exceptionReason) {
+        entry.exceptionReason = exceptionReason;
+      }
+
+      this.entries[type].push(entry);
     }
   }
 
--- a/tools/spectral/ipa/metrics/collector.js
+++ b/tools/spectral/ipa/metrics/collector.js
@@ -36,16 +36,16 @@ class Collector {
     });
   }
 
-  add(type, componentId, ruleName, exceptionReason = null ) {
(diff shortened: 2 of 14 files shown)


tools/spectral/ipa/rulesets/functions/eachCustomMethodMustBeGetOrPost.js
```diff
@@ -23,13 +24,17 @@ export default (input, opts, { path }) => {
...
+    collector.add(path, RULE_NAME, EntryType.VIOLATION);
     return ERROR_RESULT;
   }
 
   // Check for multiple valid methods
   const validMethodCount = httpMethods.filter((method) => VALID_METHODS.includes(method)).length;
 
   if (validMethodCount > 1) {
+    collector.add(path, RULE_NAME, EntryType.VIOLATION);
     return ERROR_RESULT;
```
lovisaberggren, line 36: Perhaps we can create a function for this instead, that adds to the collector and returns the error result instead, so that the collection is included by default as long as you use the same function to get the error

lovisaberggren: Though just noticed that this doesn't necessarily look the same for all rules, but would be nice if we somehow can include this automatically kind of like with the exceptions

tools/spectral/ipa/metrics/collector.js
```diff
@@ -0,0 +1,58 @@
...
+    this.fileName = 'combined.log';
+
+    process.on('exit', () => this.flushToFile());
+    process.on('SIGINT', () => {
+      this.flushToFile();
+      process.exit();
+    });
+  }
+
+  add(componentId, ruleName, type) {
```
lovisaberggren, line 39: Should we also collect the exception reasons here?

tools/spectral/ipa/metrics/collector.js
```diff
@@ -0,0 +1,58 @@
...
+      throw new Error('Use Collector.getInstance()');
+    }
+
+    this.entries = {
+      [EntryType.VIOLATION]: [],
+      [EntryType.ADOPTION]: [],
+      [EntryType.EXCEPTION]: [],
+    };
+
+    this.fileName = 'combined.log';
```
lovisaberggren, line 30: Since this file is written in the root, perhaps we should name it something more specific, like `ipa-collector-results-combined` or something like that. Also, can you add this file to the git ignore?

tools/spectral/ipa/metrics/collector.js
yelizhenden-mdb: Good catch! Yes, I'll make the changes 👍

tools/spectral/ipa/rulesets/functions/eachCustomMethodMustBeGetOrPost.js
lovisaberggren: Yeah I was thinking something like:

```
return getErrorResult(path, RULE_NAME, ERROR_RESULT)

function getErrorResult(path, ruleName, errorResult) {
  collector.add(path, ruleName, EntryType.VIOLATION);
  return errorResult;
}
```

But perhaps it's not needed, as we still need to collect the success as well in all rules, so might make more sense to keep it the same, using `collector.add()` directly

tools/spectral/ipa/rulesets/functions/utils/exceptions.js
```diff
@@ -13,3 +15,17 @@ export function hasException(object, ruleName) {
...
+ * @param object the object to evaluate
+ * @param ruleName the name of the exempted rule
+ * @param path the JSON path to the object
+ */
+export function collectException(object, ruleName, path) {
+  let exceptionReason = object[EXCEPTION_EXTENSION][ruleName];
+  if(exceptionReason) {
+    collector.add(EntryType.EXCEPTION, path, ruleName, exceptionReason);
+  }
+}
```
lovisaberggren, line 31: [Nit] I would prefer if this one is inline in `hasException()` as it needs to be called for all exceptions, or if this function would be a `getExceptionReason` and then collector.add() is called in the rule itself

yelizhenden-mdb: `hasException` currently returns a boolean to decide if something should be logged as an exception.

To make the collection process inline, we could tweak `hasException` function so that if an exception is found for the given rule, it first collects the object info as an exception and then returns true. 

Since I changed to code in a way that the other collections are also handled in the helper functions, I am happy to keep it in a consistent way. Let me know what you think

lovisaberggren: LGTM! One nit would be to have the collectException in the same utils file as the other collect functions

.gitignore
```diff
@@ -20,3 +20,4 @@
 *.vscode
 
 *.out
+/tools/spectral/ipa/ipa-collector-results-combined.log
```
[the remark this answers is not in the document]
lovisaberggren: Oh yeah it probably depends from which directory you are running the command from, whichever directory it is it will be saved there, perhaps you can add `**/*ipa-collector-results-combined.log` instead to look in any directory

lovisaberggren (approved):

PR #364, opened 2025-01-16
CLOUDP-294708: Add IPA validation to optional spec validations workflow

lovisaberggren:
## Proposed changes

Adds IPA validation to run on `optional-spec-validations` workflow. Creates an issue if the validation fails, similarly to the Postman validation.

_Jira ticket:_ [CLOUDP-294708](https://jira.mongodb.org/browse/CLOUDP-294708)

### DIFF
--- a/.github/workflows/optional-spec-validations.yml
+++ b/.github/workflows/optional-spec-validations.yml
@@ -39,6 +39,10 @@ jobs:
           name: openapi-foas-${{ inputs.env }}
           github-token: ${{ secrets.api_bot_pat }}
           run-id: ${{ github.run_id }}
+      - name: Run IPA validation
+        id: ipa-spectral-validation
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/optional-spec-validations.yml
```diff
@@ -49,7 +53,7 @@ jobs:
           make convert_to_collection
           npx -- @stoplight/spectral-cli@"${SPECTRAL_VERSION}" lint ./tmp/collection.json --ruleset=./validation/spectral.yaml
```
lovisaberggren, line 54: I wasn't sure why we run a pinned version for Postman here, passed as a variable (or where this variable is coming from). For IPA validations it will use the version from the package.json (same as the validation run on PRs/locally). We can update the Postman validation to use the same approach as IPA validation unless there is a reason for this approach here

blva (approved):

PR #386, opened 2025-01-23
CLOUDP-290416: Add workflow for releasing IPA metrics

lovisaberggren:
## Proposed changes

Adding workflow to release IPA metrics to s3.

- Uses staging creds and s3 bucket for now
- Not included in release spec workflow, this workflow will be triggered manually to verify it works as expected

Follow up:
- Add workflow to `release-spec.yml` run when we're confident it works as expected
- Update `failure-handler.yml` to also consider ipa release run result status
- Later on: When stable update to push to production s3 instead

_Jira ticket:_ [CLOUDP-290416](https://jira.mongodb.org/browse/CLOUDP-290416)

### DIFF
--- a/.github/workflows/release-IPA-metrics.yml
+++ b/.github/workflows/release-IPA-metrics.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Run Metric Collection Job
         if: ${{steps.get_previous_status.outputs.result == 'true' }}
         working-directory: ./tools/spectral/ipa/metrics/scripts
-        run: node ./tools/spectral/ipa/metrics/scripts/runMetricCollection.js ../../../../../openapi-foas.json
+        run: node runMetricCollection.js ../../../../../openapi-foas.json
 
       - name: Dump Metric Collection Job Data to S3
         if: ${{steps.get_previous_status.outputs.result == 'true' }}
--- a/.github/workflows/release-IPA-metrics.yml
+++ b/.github/workflows/release-IPA-metrics.yml
@@ -1,0 +1,74 @@
+name: IPA Validation Metrics Release
+on:
+  workflow_call:
+    secrets: # all secrets are passed explicitly in this workflow
+#      aws_access_key:
+#        required: true
+#      aws_secret_key:
+#        required: true
+#      aws_s3_bucket_prefix:
+#        required: true
+      api_bot_pat:
+        required: true
+#    inputs:
+#      env:
+#        description: 'Environment for the FOAS to use for IPA metrics collection'
+#        required: true
+#        type: string
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+jobs:
+  release-IPA-metrics:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get Previous Run Date and Status
+        id: get_previous_status
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ secrets.api_bot_pat }}
+          script: |
+            const script = require('tools/spectral/ipa/metrics/scripts/getShouldRunMetricsRelease.js')
+            const shouldRunRelease = await getShouldRunMetricsRelease({github, context})
+            return shouldRunRelease
+
+      - name: Skip Metric Collection Job
+        if: ${{steps.get_previous_status.outputs.result == 'false' }}
+        run: echo "Skipping IPA metrics release!"
+
+      - name: Setup Node
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+          cache: 'npm'
+
+      - name: Install npm dependencies
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        run: npm install
+
+      - name: Download openapi-foas
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        uses: actions/download-artifact@v4
+        with:
+          name: openapi-foas-dev  # TODO: Change to passed input env
+          github-token: ${{ secrets.api_bot_pat }}
+          run-id: ${{ github.run_id }}
+
+      - name: Run Metric Collection Job
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        working-directory: ./tools/spectral/ipa/metrics/scripts
+        run: node ./tools/spectral/ipa/metrics/scripts/runMetricCollection.js ../../../../../openapi-foas.json
+
+      - name: Dump Metric Collection Job Data to S3
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.IPA_S3_BUCKET_DW_STAGING_USERNAME }}       # TODO: Change to passed secret
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.IPA_S3_BUCKET_DW_STAGING_PASSWORD }}   # TODO: Change to passed secret
+          S3_BUCKET_PREFIX: ${{ secrets.IPA_S3_BUCKET_DW_STAGING_PREFIX }}          # TODO: Change to passed secret
+        run: node ./tools/spectral/ipa/metrics/scripts/dataDump.js openapi-foas.json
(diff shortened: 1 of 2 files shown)


.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,74 @@
...
+#        required: true
+#      aws_s3_bucket_prefix:
+#        required: true
+      api_bot_pat:
+        required: true
+#    inputs:
+#      env:
+#        description: 'Environment for the FOAS to use for IPA metrics collection'
+#        required: true
+#        type: string
```
lovisaberggren, line 17: Will update in follow-up to pass secrets from release-spec.yml

.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,74 @@
...
+  issues: write
+
+jobs:
+  release-IPA-metrics:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Get Previous Run Date and Status
```
lovisaberggren, line 30: Only run the release steps if the latest run was not today or if the last run ended in failure

.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,80 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: openapi-foas-dev  # TODO: Change to passed input env
+          github-token: ${{ secrets.api_bot_pat }}
+          run-id: ${{ github.run_id }}
+
+      - name: Run Metric Collection Job
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        working-directory: ./tools/spectral/ipa/metrics/scripts
+        run: node ./tools/spectral/ipa/metrics/scripts/runMetricCollection.js ../../../../../openapi-foas.json
```
yelizhenden-mdb, line 72: Q1: Why don't we use the downloaded `openapi-foas` from the previous step?
Q2: If we set the working directory, do we need to give the full path for the script file?

lovisaberggren: Q1: It should use the downloaded foas, the downloaded one is named openapi-foas.yaml
Q2: Good catch, did a change and forgot to change the command, it should be `node runMetricCollection.js ../../../../../openapi-foas.json`, I'll update!

.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,80 @@
...
+          github-token: ${{ secrets.api_bot_pat }}
+          run-id: ${{ github.run_id }}
+
+      - name: Run Metric Collection Job
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        working-directory: ./tools/spectral/ipa/metrics/scripts
+        run: node runMetricCollection.js ../../../../../openapi-foas.json
+
+      - name: Dump Metric Collection Job Data to S3
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
```
yelizhenden-mdb, line 75: Nit: we might use working-directory here as well

.github/workflows/release-IPA-metrics.yml
yelizhenden-mdb: I assumed it would be named something like `openapi-foas-dev`. Thanks for the info 👍

tools/spectral/ipa/metrics/scripts/getShouldRunMetricsRelease.js
```diff
@@ -0,0 +1,31 @@
...
+
+  if (runs === undefined || runs.length === 0) {
+    return true;
+  }
+
+  const previousStatus = runs[1].status;
+
+  const lastRunDate = new Date(runs[1].created_at);
+  const today = new Date();
+  const lastRunWasToday =
```
yelizhenden-mdb, line 25: Nit: you can check if two dates are equal with checking if their `toDateString()` equal to each other
`if(lastRunDate.toDateString() === today.toDateString())`

yelizhenden-mdb: Nope, it only outputs the day, month, and year ([link](https://www.geeksforgeeks.org/javascript-date-todatestring-method/))

lovisaberggren: Ah nice, missed toDateString! I'll update

yelizhenden-mdb (approved):
LGTM! Exciting! 🚀

.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,81 @@
+name: IPA Validation Metrics Release
```
andreaangiolillo, line 1: Could you add a comment at the top of this file explaining what the workflow is doing? What metrics release mean? where are they stored? By looking at the file is seems that it pushing something to s3 maybe but it is not really clear 🤔

.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,81 @@
...
+on:
+  workflow_call:
+    secrets: # all secrets are passed explicitly in this workflow
+#      aws_access_key:
+#        required: true
+#      aws_secret_key:
+#        required: true
+#      aws_s3_bucket_prefix:
+#        required: true
```
andreaangiolillo, line 10: why do you have these comments? Can we remove them?

.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,81 @@
...
+        required: true
+      IPA_S3_BUCKET_DW_STAGING_PASSWORD:
+        required: true
+      IPA_S3_BUCKET_DW_STAGING_PREFIX:
+        required: true
+#    inputs:
+#      env:
+#        description: 'Environment for the FOAS to use for IPA metrics collection'
+#        required: true
+#        type: string
```
andreaangiolillo, line 23: same here

.github/workflows/release-IPA-metrics.yml
andreaangiolillo: Could you clarify this decision? Adding a comment on the step would be beneficial to remember why we made this decision.

Why can't we run the workflow once a day automatically and file an issue on the repo in the event of a failure?

.github/workflows/release-IPA-metrics.yml
```diff
@@ -0,0 +1,81 @@
...
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        working-directory: ./tools/spectral/ipa/metrics/scripts
+        run: node runMetricCollection.js ../../../../../openapi-foas.json
+
+      - name: Dump Metric Collection Job Data to S3
+        if: ${{steps.get_previous_status.outputs.result == 'true' }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.IPA_S3_BUCKET_DW_STAGING_USERNAME }}       # TODO: Change to passed secret
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.IPA_S3_BUCKET_DW_STAGING_PASSWORD }}   # TODO: Change to passed secret
+          S3_BUCKET_PREFIX: ${{ secrets.IPA_S3_BUCKET_DW_STAGING_PREFIX }}          # TODO: Change to passed secret
```
andreaangiolillo, line 79: Can the remove the todo comments?

.github/workflows/release-IPA-metrics.yml
lovisaberggren: Sure! I can add some comments to clarify

.github/workflows/release-IPA-metrics.yml
lovisaberggren: These have been removed in this PR: https://github.com/mongodb/openapi/pull/388/files Where it's also included in the oas release workflow

.github/workflows/release-IPA-metrics.yml
lovisaberggren: Removed in this PR: https://github.com/mongodb/openapi/pull/388/files

lovisaberggren:
@andreaangiolillo 

> Could you clarify this decision? Adding a comment on the step would be beneficial to remember why we made this decision.

This was a request on the scope, to add a check before uploading the IPA metrics so that:

- If it already ran today, skip (the metrics don't need to be updated more than once a day)
- If the last run failed, run the job

I can add some comments to clarify what it does

> Why can't we run the workflow once a day automatically and file an issue on the repo in the event of a failure?

We wanted to run the workflow as part of the oas release, so that it's triggered only if there are changes detected (similarly to postman release), currently this is covered in the [follow up PR](https://github.com/mongodb/openapi/pull/388/files) which also includes IPA release in the failure-handler

PR #387, opened 2025-01-23
CLOUDP-290416: Fix and refactor IPA metrics release workflow

lovisaberggren:
## Proposed changes

Did some fixes of the IPA release workflow, added some extra error handling and logging for easier troubleshooting, and some refactoring, including splitting `release-IPA-metrics.yml` into two jobs, one for checking if the release should run, and the second for doing the release (skipped depending on the first step).

Tested the run manually on this branch, and it's passing: https://github.com/mongodb/openapi/actions/runs/12934985049

### Next steps

- Include workflow in OAS release workflow
- Pass secrets and vars from OAS release workflow to `release-IPA-metrics.yml`
- Download foas and use in `release-IPA-metrics.yml` instead of local v2.json

Later on, when we confirmed that it works as expected with the OAS release workflow, we'll change to start using the prod S3 bucket.

_Jira ticket:_ [CLOUDP-290416](https://jira.mongodb.org/browse/CLOUDP-290416)

### DIFF
--- a/.github/workflows/release-IPA-metrics.yml
+++ b/.github/workflows/release-IPA-metrics.yml
@@ -72,8 +72,7 @@ jobs:
 #          run-id: ${{ github.run_id }}
 
       - name: Run Metric Collection Job
-        working-directory: ./tools/spectral/ipa/metrics/scripts
-        run: node runMetricCollection.js ../../../../../openapi/v2.json # TODO: Change to foas in release run
+        run: node ./tools/spectral/ipa/metrics/scripts/runMetricCollection.js ${{ github.workspace }}/openapi/v2.json # TODO: Change to foas in release run
 
       - name: Dump Metric Collection Job Data to S3
         env:
--- a/.github/workflows/release-IPA-metrics.yml
+++ b/.github/workflows/release-IPA-metrics.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           github-token: ${{ secrets.api_bot_pat }}
           script: |
-            const script = require('tools/spectral/ipa/metrics/scripts/getShouldRunMetricsRelease.js')
+            const { default: getShouldRunMetricsRelease } = await import('${{ github.workspace }}/tools/spectral/ipa/metrics/scripts/getShouldRunMetricsRelease.js')
             const shouldRunRelease = await getShouldRunMetricsRelease({github, context})
             return shouldRunRelease
 
--- a/.github/workflows/release-IPA-metrics.yml
+++ b/.github/workflows/release-IPA-metrics.yml
@@ -72,7 +72,8 @@ jobs:
 #          run-id: ${{ github.run_id }}
 
       - name: Run Metric Collection Job
-        run: node ./tools/spectral/ipa/metrics/scripts/runMetricCollection.js ${{ github.workspace }}/openapi/v2.json # TODO: Change to foas in release run
+        working-directory: ./tools/spectral/ipa/metrics/scripts
+        run: node runMetricCollection.js ../../../../../openapi/v2.json # TODO: Change to foas in release run
 
       - name: Dump Metric Collection Job Data to S3
         env:
--- a/.github/workflows/release-IPA-metrics.yml
+++ b/.github/workflows/release-IPA-metrics.yml
@@ -64,16 +64,16 @@ jobs:
       - name: Install npm dependencies
         run: npm install
 
-      - name: Download openapi-foas
-        uses: actions/download-artifact@v4
-        with:
-          name: openapi-foas-dev  # TODO: Change to passed input env
-          github-token: ${{ secrets.api_bot_pat }}
-          run-id: ${{ github.run_id }}
+#      - name: Download openapi-foas
+#        uses: actions/download-artifact@v4
+#        with:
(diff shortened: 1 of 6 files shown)


.github/workflows/release-IPA-metrics.yml
```diff
@@ -39,43 +42,46 @@ jobs:
...
         run: npm install
 
-      - name: Download openapi-foas
-        if: ${{steps.get_previous_status.outputs.result == 'true' }}
-        uses: actions/download-artifact@v4
-        with:
-          name: openapi-foas-dev  # TODO: Change to passed input env
-          github-token: ${{ secrets.api_bot_pat }}
-          run-id: ${{ github.run_id }}
+#      - name: Download openapi-foas
```
yelizhenden-mdb, line 70: Q: What is the reason for commenting them out?

tools/spectral/ipa/metrics/scripts/dataDump.js
```diff
@@ -3,6 +3,14 @@ import { uploadMetricCollectionDataToS3 } from '../metricS3Upload.js';
...
 
-uploadMetricCollectionDataToS3(filePath)
-  .then(() => console.log('Data dump to S3 completed successfully.'))
-  .catch((error) => console.error(error.message));
+const response = await uploadMetricCollectionDataToS3(filePath).catch((error) => {
+  console.error(error.message);
+  process.exit(1);
+});
+
+if (!response) {
```
yelizhenden-mdb, line 11: Q: This if check here might be redundant. I could not think of any case `uploadMetricCollectionDataToS3` will return `null` or `undefined`.

tools/spectral/ipa/metrics/scripts/getShouldRunMetricsRelease.js
```diff
@@ -1,27 +1,27 @@
 // Used in .github/workflows/release-IPA-metrics.yml
 export default async function getShouldRunMetricsRelease({ github, context }) {
-  const response = await github.actions.listWorkflowRuns({
+  const response = await github.rest.actions.listWorkflowRuns({
     owner: context.repo.owner,
     repo: context.repo.repo,
-    workflow_id: context.workflow,
+    workflow_id: 'release-IPA-metrics.yml',
```
yelizhenden-mdb, line 6: Q: Why do we need to specify the workflow name?

.github/workflows/release-IPA-metrics.yml
lovisaberggren: I realized that when running the workflow independently, the foas isn't generated and uploaded first, so it can't be found here, so for now I'm using the v2 spec in the repo, but when moving the workflow into the release oas workflow, we can use this again as the foas is generated [here](https://github.com/mongodb/openapi/blob/1118fdce83599da830ef9abf6e45c55c90eaf4bc/.github/workflows/generate-openapi.yml#L76) in an earlier step

tools/spectral/ipa/metrics/scripts/dataDump.js
lovisaberggren: I added this check because if the [table creation](https://github.com/mongodb/openapi/pull/387/files#diff-fa3c82515ec5fc78dfb0ac758d81e085038fa6e95f71c208fbe2a7fd1899bfdcR19) fails due to the json not being iterable, the function will just return undefined (I think it's something with the library, imo it should error but it doesn't), I had this issue before when passing the wrong file as the collection data, which just resulted in undefined being returned, but the script should fail in this instance, so it's an extra check to make sure we actually get some result in the end

tools/spectral/ipa/metrics/scripts/getShouldRunMetricsRelease.js
lovisaberggren: When fetching the previous runs from the api, you use the workflow name or ID to specify which workflow you want to get the previous runs for

lovisaberggren: https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow

tools/spectral/ipa/metrics/scripts/dataDump.js
yelizhenden-mdb: So, this code block does not return an error somehow, am I right?

```
const table = tableFromJSON(metricsCollectionData);
  if (table === undefined) {
    throw new Error('Unable to transform metrics collection data to table');
  }
```

Sure, just wanted to be sure if it is not redundant 👍

tools/spectral/ipa/metrics/scripts/getShouldRunMetricsRelease.js
yelizhenden-mdb: So, it is not enough to get the workflow from context but we should give the name explicitly. Good find 👍

tools/spectral/ipa/metrics/scripts/dataDump.js
lovisaberggren: Yeah at least not for this particular case, adding these extra checks just as a precaution in case something isn't working as expected 👍

yelizhenden-mdb (approved):
LGTM!

PR #388, opened 2025-01-24
CLOUDP-296431: Add IPA metrics release to OAS release workflow

lovisaberggren:
## Proposed changes

- Include workflow in OAS release workflow
- Pass secrets and vars from OAS release workflow to `release-IPA-metrics.yml`
- Download FOAS and use in `release-IPA-metrics.yml` instead of local `v2.json`
- Include IPA release workflow in `failure-handler.yml`

_Jira ticket:_ [CLOUDP-296431](https://jira.mongodb.org/browse/CLOUDP-296431)

Will follow up and update to prod credentials and bucket in this ticket: [CLOUDP-296432](https://jira.mongodb.org/browse/CLOUDP-296432)

PR #390, opened 2025-01-27
CLOUDP-296909: Add some comments to the IPA release workflow

lovisaberggren:
## Proposed changes

Adds some comments describing the IPA release workflow and what it does. Addresses comments in [previous PR](https://github.com/mongodb/openapi/pull/386)

_Jira ticket:_ [CLOUDP-296909](https://jira.mongodb.org/browse/CLOUDP-296909)

### DIFF
--- a/.github/workflows/release-IPA-metrics.yml
+++ b/.github/workflows/release-IPA-metrics.yml
@@ -1,3 +1,5 @@
+# This workflow will generate metrics about the IPA validation on the current OAS
+# (number of adoptions, violations and exceptions) and upload the data to an S3 bucket.
 name: IPA Validation Metrics Release
 on:
   workflow_call:
@@ -21,6 +23,8 @@ permissions:
   issues: write
 
 jobs:
+  # Checks the last workflow run, and decides whether the IPA release metrics should be uploaded or
+  # skipped. The release-IPA-metrics job only runs if the last release was not today and did not fail.
   pre-IPA-metrics-release-checks:
     name: IPA Metrics Release Pre-Checks
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release-IPA-metrics.yml
```diff
@@ -21,6 +23,8 @@ permissions:
   issues: write
 
 jobs:
+  # Checks the last workflow run, and decides whether the IPA release metrics should be uploaded or
+  # skipped. The release-IPA-metrics job only runs if the last release was not today and did not fail.
   pre-IPA-metrics-release-checks:
```
andreaangiolillo, line 28: Do you have a Jira ticket where you plan to run the workflow as part of the release process? Btw, you could just update this workflow to run automatically when a specific file is updated on the master branch

```yaml
on:
  push:
    paths:
      - openapi/v2.json
```

andreaangiolillo (approved):
LGTM

.github/workflows/release-IPA-metrics.yml
lovisaberggren: We have already added it here: https://github.com/mongodb/openapi/pull/388 

> Btw, you could just update this workflow to run automatically when a specific file is updated on the master branch

That could be an option, though this file corresponds to the prod oas right? Since we wanted to get the metrics from the dev oas I think we need to do it during the dev oas release

andreaangiolillo: > We have already added it here
Ah okay, I missed it. I am still of the opinion that you don't need to add it to the release workflow as this makes it more complex🤔  IMO the easiest approach would be to run the workflow once a day (end of the day) and file an issue on the repo if the push fails. The Foas Release process is getting more and more complex + this approach forces us to run this workflow multiple times per day (~every hour if the spec has changed on master) while we just need to run it once.  

No need to address anything here, it is more me thinking loud.

lovisaberggren: Yeah I see your point, it's something we can consider as an improvement perhaps during the new epic IPA: validations applied, or as a follow up during a non-epic week, CC: @yelizhenden-mdb

PR #391, opened 2025-01-28
task: [spectral] make description check deep prop

gssbzn:
## Proposed changes

We have a new addition to the openapi spec that's using deeply nested properties and this linter is not catching those

I tested the changes locally to make sure it now works


andreaangiolillo (approved):
LGTM

PR #392, opened 2025-01-28
CLOUDP-297242: Fix parquet file creation for IPA metrics

lovisaberggren:
## Proposed changes

The parquet file previously created in IPA metrics collection wasn't a valid parquet file, but an arrow file. This PR adds [parquet-wasm](https://github.com/kylebarron/parquet-wasm) which can coverts an arrow table in memory to parquet which is then written to file. The file is compressed with gzip.

Tested locally and confirmed the created parquet and the dumped parquet in s3 are readable as parquet.

_Jira ticket:_ [CLOUDP-297242](https://jira.mongodb.org/browse/CLOUDP-297242)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,6 @@
 
 *.out
 **/*ipa-collector-results-combined.log
-**/*metric-collection-results.json
+**/*metric-collection-results.parquet
 **/*spectral-output.txt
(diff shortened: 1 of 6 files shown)


yelizhenden-mdb (approved):
LGTM ❤️

PR #427, opened 2025-02-14
CLOUDP-298233: use JS instead of cjs

wtrocki:
## Proposed changes

CLOUDP-298233

Make transformation to use js. Added nodejs runtime to tools as bonus.


tools/.tool-versions
```diff
@@ -1 +1,2 @@
 golang 1.23.0
+nodejs 22.11.0
```
lovisaberggren, line 2: I think we're using v20 in the GH actions workflows, could you set those to 22 or lower this one to 20 to match?

wtrocki:
```suggestion
nodejs 20.9.0
```

lovisaberggren (approved):

gssbzn:
any reason why?

wtrocki:
@gssbzn You mean node versions?

gssbzn:
> @gssbzn You mean node versions?

why move from commonjs to module js?

tools/postman/scripts/transform-postman.js
gssbzn, line 1: node uses commonjs by default should this be `.mjs`

wtrocki:
I introduced it before but we agreed with @lovisaberggren to use esm across all the repository.

tools/postman/scripts/transform-postman.js
lovisaberggren: As I understand it esm is the standard nowadays

wtrocki: We have all files as js (which should be fine) using (mjs) import format.
This PR makes single change to align to that format. 

If we want to change extension or import format let me know. Can do it across repo comprehensively.

tools/.tool-versions
```diff
@@ -1 +1,2 @@
 golang 1.23.0
+nodejs 20.9.0
```
lovisaberggren, line 2: Looked at the package json reference, we can specify there if we want: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines

wtrocki: Yes. That is very common practice. Adding it.

gssbzn: adding it to the package.json just produces warnings

lovisaberggren (approved):

PR #429, opened 2025-02-14
Add README for /tools/ipa

lovisaberggren:
Adding overview README for `/tools/ipa` about what it is, how to run, testing etc.

### DIFF
--- a/tools/spectral/ipa/README.md
+++ b/tools/spectral/ipa/README.md
@@ -1,0 +1,65 @@
+# IPA Validation
+
+The IPA validation uses [Spectral](https://docs.stoplight.io/docs/spectral/9ffa04e052cc1-spectral-cli) to validate the [MongoDB Atlas Admin API OpenAPI Specification](https://github.com/mongodb/openapi/tree/main/openapi). The rules cover MongoDB best-practices for API design.
+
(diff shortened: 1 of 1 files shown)


tools/spectral/ipa/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+The IPA validation uses [Spectral](https://docs.stoplight.io/docs/spectral/9ffa04e052cc1-spectral-cli) to validate the [MongoDB Atlas Admin API OpenAPI Specification](https://github.com/mongodb/openapi/tree/main/openapi). The rules cover MongoDB best-practices for API design.
+
+**Note:** For MongoDB engineers, please review https://go/ipa-validation-internal-wiki for additional information.
+
+## Running Locally
+
+### Prerequisites
+
+- Node.js >= v20
```
wtrocki, line 11: FUN fact: That is why package.json on ipa would be better because it can be enforced in package.json etc.

wtrocki (approved):
Superb result! all clear

tools/spectral/ipa/README.md
lovisaberggren: We could put it in package.json now as well: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#engines

PR #432, opened 2025-02-18
CLOUDP-297223: Add support for private and public preview spec generation

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-297223

- Generates private and public specs that contain only the target endpoints

## Checklist

### Changes to Spectral


tools/cli/internal/apiversion/version.go
```diff
@@ -103,6 +104,22 @@ func WithContent(contentType string) Option {
 	}
 }
 
+// WithFullContent returns an Option to generate a new APIVersion given the contentType and contentValue.
+func WithFullContent(contentType string, contentValue *openapi3.MediaType) Option {
```
blva, line 108: core change: we need to consider contentValue to gather version data, since for preview and private preview we have the same content type atlas.vnd.preview+json, we can only differentiate them based on the `x-xgen-preview` extension

tools/cli/internal/apiversion/version.go
```diff
@@ -150,9 +167,17 @@ func (v *APIVersion) IsPreview() bool {
 	return IsPreviewSabilityLevel(v.version)
 }
 
+func (v *APIVersion) IsPrivatePreview() bool {
+	return strings.Contains(v.version, PrivatePreviewStabilityLevel)
```
jeroenvervaeke, line 171: Why don't you call `IsPreviewSabilityLevel(v.version)` here?

jeroenvervaeke (commented):
LGTM with one suggestion

jeroenvervaeke (approved):

tools/cli/internal/apiversion/version.go
blva: IsPreviewSabilityLevel returns true to either Public or Private previews so we need to check for private preview here, but open to other suggestions

tools/cli/internal/cli/split/split_test.go
```diff
@@ -34,6 +38,83 @@ func TestSuccessfulSplit_Run(t *testing.T) {
 	}
 }
 
+func TestSplitPublicPreviewRun(t *testing.T) {
```
andreaangiolillo, line 41: I use the command locally to generate the spec and I noticed that the private preview file has several components that I think are not needed. Is this expected? 
[test-private-preview-info-resource.json](https://github.com/user-attachments/files/18873133/test-private-preview-info-resource.json)

I am wondering if this is okay and will be fixed once we add the filter to remove the unused view 🤔

tools/cli/internal/apiversion/version.go
```diff
@@ -247,3 +291,80 @@ func Sort(versions []*APIVersion) {
 		}
 	}
 }
+
+// GetPreviewVersionName returns the preview version name.
+func GetPreviewVersionName(contentTypeValue *openapi3.MediaType) (name string, err error) {
```
andreaangiolillo, line 296: Does this need to be public? 👀 
```suggestion
func getPreviewVersionName(contentTypeValue *openapi3.MediaType) (name string, err error) {
```

andreaangiolillo (approved):
LGTM. Great work

tools/cli/internal/apiversion/version.go
blva: I think yes cause we use it in the versions package

tools/cli/internal/cli/split/split_test.go
blva: Yes, we should cover removing unused stuff after split together, for now we're validating we're filtering out the paths, which is the most important resource

PR #443, opened 2025-02-21
CLOUDP-302126: update oasdiff

andreaangiolillo:
## Proposed changes

oasdif v1.10.26 updated the logic to consider adding schema to oneOf as breaking: https://github.com/Tufin/oasdiff/pull/611

This PR updates our changelog files to reflect this new behaviour and update the dependencies to the latest/
_Jira ticket:_ CLOUDP-302126


blva (approved):
LGTM! Can we have a follow-up item to know if we also consider it a breaking change or if we should ignore it?

andreaangiolillo:
ticket: https://jira.mongodb.org/browse/CLOUDP-302138. I will follow up on slack

PR #444, opened 2025-02-21
CLOUDP-288870: Add a ⭐️ to the most recent version of the Admin API Postman collection

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-288870

The script is changed as:
- The new collection name will start with ⭐️
- If the collection is created in the workspace first time, it will check if a collection with a star icon already exists. If so, it will modify the name removing the star icon

## Testing

I've run the commands against an empty workspace, and it created the collection with the star in the name.

Screenshot 2025-02-21 at 17 50 01

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/postman/scripts/transform-for-api.sh
+++ b/tools/postman/scripts/transform-for-api.sh
@@ -50,7 +50,7 @@ sed 's/\\"value\\": \\"<Circular reference to #[^>"]* detected>\\"//g' intermedi
 
 echo "Updating name with version $current_api_revision"
 jq --arg api_version "$current_api_revision" \
-  '.collection.info.name = ("MongoDB Atlas Administration API " + $api_version)' \
+  '.collection.info.name = ("⭐MongoDB Atlas Administration API " + $api_version)' \
   intermediateCollectionNoCircular.json >  intermediateCollectionWithName.json
 
 echo "Adding Collection description to $DESCRIPTION_FILE"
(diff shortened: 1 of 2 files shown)


tools/postman/scripts/transform-for-api.sh
```diff
@@ -50,7 +50,7 @@ sed 's/\\"value\\": \\"<Circular reference to #[^>"]* detected>\\"//g' intermedi
 
 echo "Updating name with version $current_api_revision"
 jq --arg api_version "$current_api_revision" \
-  '.collection.info.name = ("MongoDB Atlas Administration API " + $api_version)' \
+  '.collection.info.name = ("⭐MongoDB Atlas Administration API " + $api_version)' \
```
andreaangiolillo, line 53:
```suggestion
  '.collection.info.name = ("[Latest] MongoDB Atlas Administration API " + $api_version)' \
```
I am wondering why we went for the ⭐  . Isn't clearer to use Latest instead?

wtrocki (approved):
@yelizhenden-mdb We need to make change to js transformer as well, but happy to do it myself as I have introduced duplication

tools/postman/scripts/transform-for-api.sh
yelizhenden-mdb: The [discussion](https://mongodb.slack.com/archives/C07D6N6ECQP/p1733738821405619) started with using ⭐. I'm fine with changing it to [Latest]. Has there been a product decision on this? cc @matt-condon

wtrocki: @yelizhenden-mdb Go ahead and use latest as star might not be universally recognizable across the globe as "first"/"latest".

wtrocki:
Creating follow up PR.

PR #461, opened 2025-02-26
CLOUDP-304430: allow multiple formats

blva:
## Proposed changes

_Jira ticket:_ [CLOUDP-304430](https://jira.mongodb.org/browse/CLOUDP-304430)

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/cli/internal/openapi/file.go
+++ b/tools/cli/internal/openapi/file.go
@@ -29,6 +29,7 @@ import (
 const (
 	JSON    = "json"
 	YAML    = "yaml"
+	ALL     = "all"
 	DotYAML = ".yaml"
 	DotJSON = ".json"
 )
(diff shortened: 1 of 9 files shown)


tools/cli/internal/openapi/file.go
```diff
@@ -29,27 +29,34 @@ import (
 const (
 	JSON    = "json"
 	YAML    = "yaml"
+	ALL     = "all"
```
andreaangiolillo, line 32: [No need to address the comment]. Something to consider in the future is tohave the --format flag as a list so that we can provide multiple formats. The drawback of having ALL are:

- it is not clear what ALL traslate to by looking at the command but we force users to run --help
- if we add support for another format (not likely), ALL will start generating other format too 🤔 

I think is fine to have all since we are the user of this cli and it is not likely we will support other formats

andreaangiolillo (approved):
LGTM

tools/cli/internal/openapi/file.go
blva: ack, if we add more formats we can defintely revisit! or keep supporitng the all to generate all formats

PR #462, opened 2025-02-26
CLOUDP-271991: IPA-104: Validate for Get methods the response is 200

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-271991

Implements a Spectral rule to validate if Get method (single resource) has 200 OK status code and no other 2xx responses
If it is a custom method or collection resource, the rule will be bypassed.

## Checklist

### Changes to Spectral


tools/spectral/ipa/rulesets/IPA-104.yaml
```diff
@@ -21,3 +22,11 @@ rules:
     given: '$.paths[*].get'
     then:
       function: 'getMethodReturnsSingleResource'
+
+  xgen-IPA-104-get-method-response-code-is-200-OK:
```
lovisaberggren, line 26:
```suggestion
  xgen-IPA-104-get-method-response-code-is-200:
```
[Nit] If we want it a bit shorter, I think just "200" is enough

tools/spectral/ipa/rulesets/functions/getResponseCodeShouldBe200OK.js
```diff
@@ -0,0 +1,29 @@
...
+const ERROR_MESSAGE =
+  'The Get method must return a 200 OK response. This method either lacks a 200 OK response or defines a different 2xx status code.';
+
+export default (input, _, { path }) => {
+  if (hasException(input, RULE_NAME)) {
+    collectException(input, RULE_NAME, path);
+    return;
+  }
+
+  if (input['responses']) {
```
lovisaberggren, line 14: Something to consider here is that this test will cover IPA 105 List as well, since it's also a HTTP `GET` method, which may skew the metrics, and will cause duplicate validation if we are planning to implement the same check for IPA 105.

Also, this covers custom methods too, though IIRC we don't define that custom method `GET` should return 200. Maybe we should introduce a guideline for this on custom methods, or not do the validation if we think there is a reason for custom gets to return a 201 for example (though IMO it should be a custom post in that case)

yelizhenden-mdb: I modified the code as "If it is a custom method or collection resource (not a child with current glossary), the rule will be bypassed". Let me know what you think

lovisaberggren: Thanks!

lovisaberggren (approved):

PR #463, opened 2025-02-27
CLOUDP-302984: IPA rule Get methods should return data for a single resource

lovisaberggren:
## Proposed changes

Adds new IPA rule `xgen-IPA-104-get-method-returns-single-resource` to check that get operations don't return paginated or array responses. Applied to `GET` for standard resources and singleton resources.

_Jira ticket:_ [CLOUDP-302984](https://jira.mongodb.org/browse/CLOUDP-302984)

### Testing

- Added unit test for rule
- Checked current offenders, there are 23 violations currently

See all violations: https://github.com/mongodb/openapi/pull/463/checks?check_run_id=37928281663

Example violation:

```
warning  xgen-IPA-104-get-method-returns-single-resource  Get methods should return data for a single resource. This method returns an array or a paginated response. http://go/ipa/104  paths./api/atlas/v2/groups/{groupId}/clusters/{clusterName}/fts/indexes/{databaseName}/{collectionName}.get.responses[200].content.application/vnd.atlas.2023-01-01+json
```

### DIFF
--- a/tools/spectral/ipa/__tests__/getResourceIsNotPaginated.test.js
+++ b/tools/spectral/ipa/__tests__/getResourceIsNotPaginated.test.js
@@ -1,0 +1,298 @@
+import testRule from './__helpers__/testRule';
+import { DiagnosticSeverity } from '@stoplight/types';
+
+const componentSchemas = {
+  schemas: {
+    Schema: {
+      properties: {
+        exampleProperty: {
+          type: 'string',
+        },
+      },
+    },
+    PaginatedSchema: {
+      type: 'array',
+    },
+    ArraySchema: {
+      properties: {
+        results: {
+          type: 'array',
+        },
+      },
+    },
+  },
+};
+
+testRule('xgen-IPA-104-GET-resource-not-paginated', [
+  {
+    name: 'valid resources',
+    document: {
+      paths: {
+        '/resource': {
+          get: {
+            responses: {
+              200: {
+                content: {
+                  'application/vnd.atlas.2024-08-05+json': {
+                    schema: {
+                      $ref: '#/components/schemas/PaginatedSchema',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+        '/resource/{id}': {
+          get: {
+            responses: {
+              200: {
+                content: {
+                  'application/vnd.atlas.2024-08-05+json': {
+                    schema: {
+                      $ref: '#/components/schemas/Schema',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+        '/resource/{id}:getAllThings': {
+          get: {
+            responses: {
+              200: {
+                content: {
+                  'application/vnd.atlas.2024-08-05+json': {
+                    schema: {
+                      $ref: '#/components/schemas/ArraySchema',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+        '/singleton': {
+          get: {
+            responses: {
+              200: {
+                content: {
+                  'application/vnd.atlas.2024-08-05+json': {
+                    schema: {
+                      $ref: '#/components/schemas/Schema',
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+      components: componentSchemas,
+    },
+    errors: [],
+  },
+  {
+    name: 'invalid resources',
+    document: {
+      paths: {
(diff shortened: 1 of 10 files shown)


tools/spectral/ipa/__tests__/getResourceIsNotPaginated.test.js
```diff
@@ -0,0 +1,298 @@
...
+      },
+      components: componentSchemas,
+    },
+    errors: [],
+  },
+  {
+    name: 'invalid resources',
+    document: {
+      paths: {
+        '/arrayResource': {
```
yelizhenden-mdb, line 100: Do we need this one among invalid resources? It seems valid

tools/spectral/ipa/__tests__/getResourceIsNotPaginated.test.js
```diff
@@ -0,0 +1,298 @@
...
+        ],
+        severity: DiagnosticSeverity.Warning,
+      },
+    ],
+  },
+  {
+    name: 'invalid resource with exception',
+    document: {
+      paths: {
+        '/arrayResource': {
```
yelizhenden-mdb, line 260: Same question with above. It is not an invalid resource with exception

tools/spectral/ipa/rulesets/IPA-104.yaml
```diff
@@ -13,3 +14,10 @@ rules:
     then:
       field: '@key'
       function: 'eachResourceHasGetMethod'
+  xgen-IPA-104-GET-resource-not-paginated:
```
yelizhenden-mdb, line 17: Get-method-returns-single-resource might be a better name, because we expect the resource not paginated and not array

tools/spectral/ipa/rulesets/IPA-104.yaml
```diff
@@ -13,3 +14,10 @@ rules:
...
+    description: 'The purpose of the get method is to return data from a single resource. http://go/ipa/104'
+    message: '{{error}} http://go/ipa/104'
+    severity: warn
+    given: '$.paths[*].get'
+    then:
+      function: 'getResourceIsNotPaginated'
```
yelizhenden-mdb, line 23: Similarly, function name might change

tools/spectral/ipa/__tests__/getResourceIsNotPaginated.test.js
lovisaberggren: Had it here to make sure the error is on the child and not the parent, and to demonstrate a parent+child resource, though this should be covered in the valid resources test case, so I can remove it

tools/spectral/ipa/rulesets/IPA-104.yaml
lovisaberggren: Makes sense, thanks! I'll update

tools/spectral/ipa/__tests__/getResourceIsNotPaginated.test.js
yelizhenden-mdb: Yes, the point I am trying to make is we can only keep the invalid ones in this test case

yelizhenden-mdb (approved):
LGTM

tools/spectral/ipa/rulesets/functions/getMethodReturnsSingleResource.js
```diff
@@ -0,0 +1,39 @@
...
+  const errors = [];
+
+  const responseSchemas = getAllSuccessfulResponseSchemas(input);
+  responseSchemas.forEach(({ schemaPath, schema }) => {
+    const fullPath = path.concat(schemaPath);
+    const responseObject = resolveObject(oas, fullPath);
+
+    if (hasException(responseObject, RULE_NAME)) {
+      collectException(responseObject, RULE_NAME, fullPath);
+    } else if (schemaIsPaginated(schema) || schemaIsArray(schema)) {
```
wtrocki, line 30: Question - when we have PR level job does this validator runs agains dev or production schema?

lovisaberggren: It will be for the dev schema

wtrocki: Perfect!

PR #466, opened 2025-02-28
fix: request body for preview is empty

andreaangiolillo:
## Proposed changes

Ticket: [CLOUDP-303647](https://jira.mongodb.org/browse/CLOUDP-303647)
## Bug
While migrating the API to preview in this PR https://github.com/10gen/mms/pull/122517, I noticed that the request body for the endpoint was not being populated. After reproducing the issue locally, I found the root cause: we use `time.Now()` with the preview API, and when comparing the content type for preview using `time.GreaterThan`, the comparison always returned `true`. This happened because `time.GreaterThan` was factoring in minutes and seconds, even though both versions were set to preview.

## Fix
I updated the logic to use the same date as we have in Atlas so that we don't set HH:MM:SS.
I regenerated the preview spec locally and confirmed that the request body was correctly populated.

### DIFF
--- a/tools/cli/internal/openapi/filter/versioning.go
+++ b/tools/cli/internal/openapi/filter/versioning.go
@@ -22,7 +22,7 @@ import (
 	"github.com/mongodb/openapi/tools/cli/internal/apiversion"
 )
 
-// VersioningFilter is a filter that modifies the OpenAPI spec by removing operations and responses
+// Filter: VersioningFilter is a filter that modifies the OpenAPI spec by removing operations and responses
 // that are not supported by the target version.
 type VersioningFilter struct {
 	oas      *openapi3.T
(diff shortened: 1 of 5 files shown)


tools/cli/internal/openapi/filter/versioning.go
```diff
@@ -21,7 +22,7 @@ import (
 	"github.com/mongodb/openapi/tools/cli/internal/apiversion"
 )
 
-// Filter: VersioningFilter is a filter that modifies the OpenAPI spec by removing operations and responses
```
blva, line 24: `// Filter: ` notation is used so that we can auto gen all the filters

blva (approved):
nice! a great benefit of moving over to foascli, now our feedback loop is much quicker to fix issues

tools/cli/internal/openapi/filter/versioning.go
andreaangiolillo: ah I was getting a warning in my IDE that this is not golang standard. I will revert the changes

andreaangiolillo: I reverted my changes. thanks!

PR #469, opened 2025-03-03
CLOUDP-303615: IPA - Update check if resource is singleton

lovisaberggren:
## Proposed changes

After discussions we have decided to slightly modify the check for if a resource is a "standard" resource or a singleton resource. For most resources, they will still be treated the same, but some resources will not be considered singleton anymore.

- The concept of a "standard" resource is gone, since the IPAs don't really reflect this
- A resource is either singleton or not, and if it's not, we would expect it to have the standard methods (and optionally custom methods)
- Modified the singleton resource check to also look at if there is a post in some cases, the logic is explained in [this diagram](https://lucid.app/lucidchart/d94b9cf7-f040-403a-b74d-8c475b021054/edit?view_items=TVhYQnfmzZ2~&invitationId=inv_1d36f12e-4026-423b-8bc8-4dc2dfc65eb3)
- Added checks for if a path refers to "collection of resources" or "a single resource" instead of "child" for example, since it actually means a nested resource. Helps us correctly evaluate paths/resources, and allows us to disregard some weird paths that probably should be custom methods if implemented today (for example `/resource/{id}/something/verb`)

### Tests

- Updated tests for `resourceEvaluation` for checking for singleton resource, collection path or single path
- Updated tests

## Follow up

- Add/remove exceptions that are not needed anymore or that are required now

_Jira ticket:_ [CLOUDP-303615](https://jira.mongodb.org/browse/CLOUDP-303615)


tools/spectral/ipa/__tests__/singletonHasNoId.test.js
```diff
@@ -47,7 +47,7 @@ testRule('xgen-IPA-113-singleton-must-not-have-id', [
           patch: {},
           delete: {},
         },
-        '/singleton1': {
+        '/standard/{exampleId}/singleton1': {
```
yelizhenden-mdb, line 50: I would prefer a usage similar to `/resource/{id}/singleton` used before. Seeing standard and singleton at the same path is a bit confusing

tools/spectral/ipa/rulesets/functions/eachResourceHasGetMethod.js
```diff
@@ -31,6 +31,9 @@ export default (input, _, { path, documentInventory }) => {
       return collectAndReturnViolation(path, RULE_NAME, ERROR_MESSAGE);
     }
   } else if (isStandardResource(resourcePaths)) {
+    if (resourcePaths.length === 1) {
```
yelizhenden-mdb, line 34: Isn't this use case captured under `isSingletonResource`?

yelizhenden-mdb: Now I see the implementation is changed for singletonResource, so this comment is irrelevant at the moment

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
```diff
@@ -5,84 +5,186 @@ import {
   isStandardResource,
 } from '../../rulesets/functions/utils/resourceEvaluation';
 
-const standardResourcePaths = ['/standard', '/standard/{id}'];
+const standardResourcePaths = ['/standard', '/standard/{exampleId}'];
 
 const nestedStandardResourcePaths = ['/standard/{exampleId}/nested', '/standard/{exampleId}/nested/{exampleId}'];
 
-const standardResourceWithCustomPaths = ['/customStandard', '/customStandard/{id}', '/customStandard:method'];
+const standardResourceWithCustomPaths = ['/customStandard', '/customStandard/{exampleId}', '/customStandard:method'];
```
yelizhenden-mdb, line 12: Should we include `/customStandard/{exampleId}:method` as well?

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
```diff
@@ -5,84 +5,186 @@ import {
...
 
-const singletonResourcePaths = ['/singleton'];
+const standardResourceMissingSubPath = ['/standardMissingSub'];
 
-const singletonResourceWithCustomPaths = ['/customSingleton', '/customSingleton:method'];
+const standardResourceMissingSubPathInvalidFormat = ['/standard/nestedStandardMissingSub'];
 
-const nestedSingletonResourcePaths = ['/standard/{exampleId}/nestedSingleton'];
+const standardResourcePathsInvalidFormat = ['/resource1/resource2', '/resource1/resource2/{exampleId}'];
```
yelizhenden-mdb, line 18: Do we have such cases as `/resource1/resource2/{exampleId}/{exampleId2}`?

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
```diff
@@ -5,84 +5,186 @@ import {
...
-      expect(getResourcePaths('/singleton', allPaths)).toEqual(singletonResourcePaths);
-      expect(getResourcePaths('/customSingleton', allPaths)).toEqual(singletonResourceWithCustomPaths);
-      expect(getResourcePaths('/standard/{exampleId}/nestedSingleton', allPaths)).toEqual(nestedSingletonResourcePaths);
-    });
-  });
-  describe('isStandardResource', () => {
-    it('returns true for a standard resource', () => {
-      expect(isStandardResource(standardResourcePaths)).toBe(true);
-    });
+const singletonResourcePaths = ['/standard/{exampleId}/singleton'];
```
yelizhenden-mdb, line 20: Can we have such case `/resource`, ` /resource/{id}` and `/resource/{id}/singleton`, and we should verify two first two are standard but the last one is singleton because there is no `/resource/{id}/singleton/{singletonId}`. Or, is there any test case to test it I probably missed?

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
```diff
@@ -5,84 +5,186 @@ import {
...
-const standardResourcePaths = ['/standard', '/standard/{id}'];
+const standardResourcePaths = ['/standard', '/standard/{exampleId}'];
 
 const nestedStandardResourcePaths = ['/standard/{exampleId}/nested', '/standard/{exampleId}/nested/{exampleId}'];
 
-const standardResourceWithCustomPaths = ['/customStandard', '/customStandard/{id}', '/customStandard:method'];
+const standardResourceWithCustomPaths = ['/customStandard', '/customStandard/{exampleId}', '/customStandard:method'];
 
-const singletonResourcePaths = ['/singleton'];
+const standardResourceMissingSubPath = ['/standardMissingSub'];
```
yelizhenden-mdb, line 14: I don't know if we have such cases but what if there is `/resource/{id}` without `/resource`

wtrocki: That should be instant failure. I cannot imagine `cloud.mongodb.com/api/v2/{id}`

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
wtrocki: To aid disscussion we discourage nested resources:
`/resource/{id}/resource`

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
wtrocki: Also for purpose of the mapping of existing API and IPA you should assume that org and group prefixes are not existent. e.g 
```
/org/{orgId}/... 
/group/{groupId}/... 
```
 do not count

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
lovisaberggren: Sure, I'll add 👍

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
lovisaberggren: @yelizhenden-mdb The case of `/resource/{id}/{anotherId}` is a weird case as it's not a valid path format according to IPAs, so right now these are excluded from the places where we fetch all paths for a resource. We could include them but I am not sure how to handle them in related rules, for example if we are checking that a resource has a get method, do we check `/resource/{id}/{anotherId}` or only `/resource/{id}`, or both? Since it's not clear in our guidelines they are not included, but they still violate the rule about the path format which flags in the future that this shouldn't be done

> Also for purpose of the mapping of existing API and IPA you should assume that org and group prefixes are not existent. e.g

@wtrocki Why not cover them?

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
lovisaberggren: @wtrocki I don't think we explicitly discourage it, for example [IPA 101](https://docs.devprod.prod.corp.mongodb.com/ipa/101)

> Resources may have any number of sub-resources

> Can we have such case /resource,  /resource/{id} and /resource/{id}/singleton, and we should verify two first two are standard but the last one is singleton because there is no /resource/{id}/singleton/{singletonId}. Or, is there any test case to test it I probably missed?

Yes that's a valid case. The `standardResourcePaths` is the parent (`/standard/{exampleId}`), and the singleton is a child, `/standard/{exampleId}/singleton`

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
lovisaberggren: It could happen I think, though I don't think we have a rule that would cover it specifically

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
yelizhenden-mdb: My intention was to ask if there is a test case for this scenario 😄

lovisaberggren: Yes it's included in this file 👍

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
yelizhenden-mdb: I think it is pretty possible if the API only supports the Get method but not the List and Create method. I know , according to the guidelines, the resource should provide List method unless it is a singleton, but assuming the worst case here because we always assume there are ones that violate these rules

wtrocki (approved):

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
wtrocki: Clarified offline. We need to clarify them as they are unique resource. 
Separately most of the `/resource1/resource2/` use cases could be qualified as custom methods

yelizhenden-mdb: We have this example: `/api/atlas/v2/groups/{groupId}/cloudProviderAccess/{cloudProvider}/{roleId}`. As you said, this is a best effort, therefore, maybe we can have `uncategorized`, which will mean that we don't know if it is singleton/standard, single/multiple resource, and we will not apply any rules

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
wtrocki: > @wtrocki I don't think we explicitly discourage it, for example [IPA 101]
Apologies. I meant
`/resource/singleton`.

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
lovisaberggren: I think we should still apply rules to these endpoints/operations (other types of rules that are not tied to singleton/standard resources). I don't think these need to be categorized as they already are disregarded when we are fetching all paths for a resource. Meaning, currently if we would get all paths for the resource `/api/atlas/v2/groups/{groupId}/cloudProviderAccess`, `/api/atlas/v2/groups/{groupId}/cloudProviderAccess/{cloudProvider}` would be included (for example checking if there is a Get), but `/api/atlas/v2/groups/{groupId}/cloudProviderAccess/{cloudProvider}/{roleId}` would not

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
lovisaberggren: We could implement a function to get all paths for a resource based on a path like `/resource/{id}` when there is no `/resource`, as part of the List/Create IPA implementation, but that could be a separate helper that we introduce later when we need it. WDYT?

tools/spectral/ipa/__tests__/utils/resourceEvaluation.test.js
lovisaberggren: Yup, that would be considered standard instead 👍

tools/spectral/ipa/rulesets/functions/getMethodReturnsSingleResource.js
```diff
@@ -8,32 +13,42 @@ import { resolveObject } from './utils/componentUtils.js';
 const RULE_NAME = 'xgen-IPA-104-get-method-returns-single-resource';
 const ERROR_MESSAGE_STANDARD_RESOURCE =
   'Get methods should return data for a single resource. This method returns an array or a paginated response.';
+const ERROR_MESSAGE_SINGLETON_RESOURCE =
```
yelizhenden-mdb, line 16: ❤️

tools/spectral/ipa/rulesets/functions/getMethodReturnsSingleResource.js
```diff
@@ -8,32 +13,42 @@ import { resolveObject } from './utils/componentUtils.js';
...
-      collectAndReturnViolation(fullPath, RULE_NAME, ERROR_MESSAGE_STANDARD_RESOURCE);
-      errors.push({ path: fullPath, message: ERROR_MESSAGE_STANDARD_RESOURCE });
-    } else {
-      collectAdoption(fullPath, RULE_NAME);
-    }
-  });
+      if (hasException(responseObject, RULE_NAME)) {
+        collectException(responseObject, RULE_NAME, fullPath);
+      } else if (schemaIsPaginated(schema) || schemaIsArray(schema)) {
+        collectAndReturnViolation(
```
yelizhenden-mdb, line 38: When you `return collectAndReturnViolation`, it should return the error. Didn't it work that way? Is there a reason not returning here?

yelizhenden-mdb: Okay, I see now it is inside `forEach`, ignore my comment

yelizhenden-mdb (approved):
LGTM. High-value foundational work!

PR #472, opened 2025-03-04
task: spectral - minor changes in readme and dev setup

wtrocki:
Couple drivebys while I'm reviewing the codebase:

- Added missing make targets (We might want to move out of make and use NPM directly. 
- Added more structure to the main readme file pointing to the rulesets
-


tools/spectral/Makefile
```diff
@@ -26,6 +25,17 @@ lintwarn: ## Run spectral linter on foas
 	@echo "==> Running spectral linter"
 	npx -- @stoplight/spectral-cli@"${SPECTRAL_VERSION}" lint ../../openapi/v2.yaml --ruleset=.spectral.yaml --fail-severity=warn
 
+## IPA sepecific targets
```
wtrocki, line 28: Ideally we should not use make but wanted to add targets for completeness and simplicity of onboarding.

lovisaberggren: I don't think we need to add the IPA targets here since we have the package.json, this Makefile is for the legacy Spectral we have. I think I'd prefer if we move these to the package.json instead

blva: my two cents from the outside is that i'd love to see a makefile to quickly catch up since other tools also have makefiles (no strong opinion)

wtrocki: Yes. make targets are legacy and I was supprised to see them. Do you know if we use those in dowstream?

wtrocki: For now I will remove those changes and we can have follow up PR to cover removal of MAKE (likely needs more than just this PR)

lovisaberggren (approved):

PR #483, opened 2025-03-06
 CLOUDP-271997: IPA 108 - delete method schema

wtrocki:
## Proposed changes
 
 CLOUDP-271997

# Changes Overview
This first PR implements OpenAPI specification validation for IPA-108 guidelines related to DELETE operations. The implementation adds linting rules that verify DELETE methods follow REST API best practices.

# Components Implemented
Rule: xgen-IPA-108-delete-response-should-be-empty

- Validates that DELETE operations return empty responses (204 No Content)
- Ensures response objects don't include schema definitions
- Implemented in deleteMethodResponseShouldNotHaveSchema.js

### DIFF
--- a/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
+++ b/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
@@ -30,6 +30,9 @@ testRule('xgen-IPA-108-delete-response-should-be-empty', [
                   'application/vnd.atlas.2023-01-01+json': {
                     'x-xgen-version': '2023-01-01',
                   },
+                  'application/vnd.atlas.2023-03-01+json': {
+                    'x-xgen-version': '2023-01-01',
+                  },
                 },
               },
             },
@@ -62,7 +65,7 @@ testRule('xgen-IPA-108-delete-response-should-be-empty', [
       {
         code: 'xgen-IPA-108-delete-response-should-be-empty',
         message:
-          'DELETE method should return an empty response. The response should not have a schema property and reference to models. http://go/ipa/108',
+          'Error found for application/vnd.atlas.2023-01-01+json: DELETE method should return an empty response. The response should not have a schema property. http://go/ipa/108',
         path: ['paths', '/resource/{id}', 'delete'],
         severity: DiagnosticSeverity.Warning,
       },
--- a/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
+++ b/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
@@ -66,7 +66,7 @@ testRule('xgen-IPA-108-delete-response-should-be-empty', [
         code: 'xgen-IPA-108-delete-response-should-be-empty',
         message:
           'Error found for application/vnd.atlas.2023-01-01+json: DELETE method should return an empty response. The response should not have a schema property. http://go/ipa/108',
-        path: ['paths', '/resource/{id}', 'delete'],
+        path: ['paths', '/resource/{id}', 'delete', 'responses', '204'],
         severity: DiagnosticSeverity.Warning,
       },
     ],
@@ -77,13 +77,13 @@ testRule('xgen-IPA-108-delete-response-should-be-empty', [
       paths: {
         '/resource/{id}': {
           delete: {
-            'x-xgen-IPA-exception': {
-              'xgen-IPA-108-delete-response-should-be-empty': 'Legacy API',
-            },
             responses: {
               204: {
-                content: {
-                  'application/vnd.atlas.2023-01-01+json': {
+                'application/vnd.atlas.2023-01-01+json': {
+                  'x-xgen-IPA-exception': {
+                    'xgen-IPA-108-delete-response-should-be-empty': 'Legacy API',
+                  },
+                  content: {
                     schema: { type: 'object' },
                   },
                 },
--- a/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
+++ b/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
@@ -18,7 +18,7 @@ testRule('xgen-IPA-108-delete-response-should-be-empty', [
     errors: [],
   },
   {
-    name: 'valid DELETE with void 204',
+    name: 'valid DELETE with void 204 versioned',
     document: {
       paths: {
         '/resource/{id}': {
@@ -62,7 +62,7 @@ testRule('xgen-IPA-108-delete-response-should-be-empty', [
       {
         code: 'xgen-IPA-108-delete-response-should-be-empty',
         message:
-          'DELETE method should return an empty response. The response should not have a schema property and reference to models http://go/ipa/108',
+          'DELETE method should return an empty response. The response should not have a schema property and reference to models. http://go/ipa/108',
         path: ['paths', '/resource/{id}', 'delete'],
         severity: DiagnosticSeverity.Warning,
       },
--- a/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
+++ b/tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
@@ -1,0 +1,95 @@
+import testRule from './__helpers__/testRule';
+import { DiagnosticSeverity } from '@stoplight/types';
+
+testRule('xgen-IPA-108-delete-response-should-be-empty', [
+  {
+    name: 'valid DELETE with void 204',
+    document: {
+      paths: {
+        '/resource/{id}': {
+          delete: {
+            responses: {
+              204: {},
+            },
+          },
+        },
+      },
+    },
+    errors: [],
+  },
+  {
+    name: 'valid DELETE with void 204',
+    document: {
+      paths: {
+        '/resource/{id}': {
+          delete: {
+            responses: {
+              204: {
+                description: 'No Content',
+                content: {
+                  'application/vnd.atlas.2023-01-01+json': {
+                    'x-xgen-version': '2023-01-01',
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+    errors: [],
+  },
+  {
+    name: 'invalid DELETE with non-void 204',
+    document: {
+      paths: {
+        '/resource/{id}': {
+          delete: {
+            responses: {
+              204: {
+                content: {
+                  'application/vnd.atlas.2023-01-01+json': {
+                    schema: { type: 'object' },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+    errors: [
+      {
+        code: 'xgen-IPA-108-delete-response-should-be-empty',
+        message:
+          'DELETE method should return an empty response. The response should not have a schema property and reference to models http://go/ipa/108',
+        path: ['paths', '/resource/{id}', 'delete'],
+        severity: DiagnosticSeverity.Warning,
+      },
+    ],
+  },
+  {
+    name: 'valid with exception',
+    document: {
+      paths: {
+        '/resource/{id}': {
+          delete: {
+            'x-xgen-IPA-exception': {
+              'xgen-IPA-108-delete-response-should-be-empty': 'Legacy API',
+            },
+            responses: {
+              204: {
+                content: {
+                  'application/vnd.atlas.2023-01-01+json': {
+                    schema: { type: 'object' },
+                  },
+                },
+              },
+            },
+          },
+        },
+      },
+    },
+    errors: [],
+  },
+]);
--- a/tools/spectral/ipa/rulesets/IPA-108.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-108.yaml
@@ -1,0 +1,14 @@
+# IPA-108: Delete
+# http://go/ipa/108
+
+rules:
+  xgen-IPA-108-delete-response-should-be-empty:
+    description: Delete method response should not have schema reference to object http://go/ipa/108
+    message: '{{error}} http://go/ipa/108'
+    severity: warn
+    given: $.paths[*].delete
+    then:
+      function: deleteMethodResponseShouldNotHaveSchema
+
+functions:
+  - deleteMethodResponseShouldNotHaveSchema
--- a/tools/spectral/ipa/rulesets/IPA-108.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-108.yaml
@@ -3,7 +3,7 @@
 
 rules:
   xgen-IPA-108-delete-response-should-be-empty:
-    description: Delete method response should not have schema reference to object http://go/ipa/108
+    description: Delete method response should not have schema reference to object. http://go/ipa/108
     message: '{{error}} http://go/ipa/108'
     severity: warn
     given: $.paths[*].delete
--- a/tools/spectral/ipa/rulesets/IPA-108.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-108.yaml
@@ -6,7 +6,7 @@ rules:
     description: Delete method response should not have schema reference to object. http://go/ipa/108
     message: '{{error}} http://go/ipa/108'
     severity: warn
-    given: $.paths[*].delete
+    given: $.paths[*].delete.responses[204]
     then:
       function: deleteMethodResponseShouldNotHaveSchema
 
--- a/tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
+++ b/tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
@@ -1,0 +1,34 @@
+import { hasException } from './utils/exceptions.js';
+import { collectAdoption, collectAndReturnViolation, collectException } from './utils/collectionUtils.js';
+
+const RULE_NAME = 'xgen-IPA-108-delete-response-should-be-empty';
+const ERROR_MESSAGE =
+  'DELETE method should return an empty response. The response should not have a schema property and reference to models';
+
+/**
+ * Delete method should return an empty response
+ * @param {object} input - The delete operation object
+ * @param {object} _ - Unused
+ * @param {object} context - The context object containing the path
+ */
+export default (input, _, { path }) => {
+  const deleteOp = input;
+
+  if (hasException(deleteOp, RULE_NAME)) {
+    collectException(deleteOp, RULE_NAME, path);
+    return;
+  }
+
+  const responses = deleteOp.responses || {};
+  for (const [status, response] of Object.entries(responses)) {
+    if (status === '204' && response.content) {
+      for (const contentType of Object.keys(response.content)) {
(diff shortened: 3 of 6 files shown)


tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+testRule('xgen-IPA-108-delete-response-should-be-empty', [
+  {
+    name: 'valid DELETE with void 204',
+    document: {
+      paths: {
+        '/resource/{id}': {
+          delete: {
+            responses: {
+              204: {},
```
yelizhenden-mdb, line 12: How about the ones with no schema? Don't we count them as empty?

Example:
```
 "204": {
            "description": "No Content",
            "content": {
              "application/vnd.atlas.2023-01-01+json": {
                "x-xgen-version": "2023-01-01"
              }
            }
          }
```

gssbzn (approved):

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Delete method should return an empty response
+ * @param {object} input - The delete operation object
+ * @param {object} _ - Unused
+ * @param {object} context - The context object containing the path
+ */
+export default (input, _, { path }) => {
+  const deleteOp = input;
+  if (!deleteOp) return;
```
yelizhenden-mdb, line 16: Which case is the equivalent of this check? `delete` method without any fields? Is it possible?

If there is no `delete` method for the resource, the rule will not run anyway, given the rule definition

yelizhenden-mdb: Maybe returning if deleteOp.responses is null or empty would make more sense. Think of the cases where you would say this rule is adopted.

I would say if the delete method has responses and the successful response is empty, I would say the delete method has adopted this rule. This implementation we are saying that the delete method adopts the rule even if there are no responses defined

tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
wtrocki: Yes. We need it for versioning engine to function correctly.

tools/spectral/ipa/rulesets/IPA-108.yaml
```diff
@@ -0,0 +1,15 @@
+# IPA-108: Delete
+# http://go/ipa/108
+
+rules:
+  xgen-IPA-108-delete-response-should-be-empty:
+    description: Delete method response should not have schema reference to object http://go/ipa/108
```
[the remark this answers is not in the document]
lovisaberggren: [Nit]

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,35 @@
+import { hasException } from './utils/exceptions.js';
+import { collectAdoption, collectAndReturnViolation, collectException } from './utils/collectionUtils.js';
+
+const RULE_NAME = 'xgen-IPA-108-delete-response-should-be-empty';
+const ERROR_MESSAGE =
+  'DELETE method should return an empty response. The response should not have a schema property and reference to models';
```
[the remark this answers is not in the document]
lovisaberggren: [Nit]

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
wtrocki: Great points. I recommend us to write JS code that always ensures no undefined errors will happen even when inprobable to happen. That practice will allow us to avoid errors during the 
For spectral and openapi deleteOp will be always defined  (not confused by empty object).

Separately you have hit very important question of filtering rules without adoption. 
We have that in number of existing policies thus I adopted this pattern e.g:
https://github.com/mongodb/openapi/blob/55e9c04afd23265f48c81d736f69bae47664a86d/tools/spectral/ipa/rulesets/functions/eachResourceHasGetMethod.js#L16

wtrocki: TL;DR - will go ahead and remove it althought I do feel that we need to set JS code standards within the team to avoid similar conversations on PRs in the future.

tools/spectral/ipa/rulesets/IPA-108.yaml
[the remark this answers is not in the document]
wtrocki: Nice!

tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
wtrocki: Done

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Delete method should return an empty response
+ * @param {object} input - The delete operation object
+ * @param {object} _ - Unused
+ * @param {object} context - The context object containing the path
+ */
+export default (input, _, { path }) => {
+  const deleteOp = input;
+
```
yelizhenden-mdb, line 16:
```suggestion
if(!deleteOp.responses || deleteOp.responses.length === 0) {
    return; 
}
```

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,34 @@
...
+  if (hasException(deleteOp, RULE_NAME)) {
+    collectException(deleteOp, RULE_NAME, path);
+    return;
+  }
+
+  const responses = deleteOp.responses || {};
```
yelizhenden-mdb, line 22:
```suggestion
  const responses = deleteOp.responses;
```

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
wtrocki: SGTM. That will mean we want rules to have [4 states as discussed above](https://github.com/mongodb/openapi/pull/483#discussion_r1983177479):
- State of success (without reporting compilance)
- State of success with compilance
- Exemption 
- Failure

yelizhenden-mdb: Yes, we can note that we are assuming that there will be no undefined input variable 👍

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Delete method should return an empty response
+ * @param {object} input - The delete operation object
+ * @param {object} _ - Unused
+ * @param {object} context - The context object containing the path
+ */
+export default (input, _, { path }) => {
+  const deleteOp = input;
+if(!deleteOp.responses || deleteOp.responses.length === 0) {
```
yelizhenden-mdb, line 16: Probably `prettier` will complain about this :D

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,36 @@
...
+    return; 
+}
+  if (hasException(deleteOp, RULE_NAME)) {
+    collectException(deleteOp, RULE_NAME, path);
+    return;
+  }
+
+  const responses = deleteOp.responses;
+  for (const [status, response] of Object.entries(responses)) {
+    if (status === '204' && response.content) {
```
yelizhenden-mdb, line 26: If we only check for status == 204, do we need the for loop on the previous line?

wtrocki: Yes. Covered in [spec](https://docs.google.com/document/d/1sMwvhyzUaAa8Y56WXHHj44MVoEaedLsQh45K0ZwnKE4/edit?tab=t.ptrwft2w5x35#bookmark=id.e5kyk2qg3lp7) based on team discussion we had

wtrocki: We cannot make strong assumptions to the API format e.g:


Covered that in spec: 

https://docs.google.com/document/d/1sMwvhyzUaAa8Y56WXHHj44MVoEaedLsQh45K0ZwnKE4/edit?tab=t.ptrwft2w5x35#bookmark=id.e5kyk2qg3lp7

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,36 @@
...
+      for (const contentType of Object.keys(response.content)) {
+        if (response.content[contentType].schema) {
+          return collectAndReturnViolation(path, RULE_NAME, ERROR_MESSAGE);
+        }
+      }
+    }
+  }
+
+  collectAdoption(path, RULE_NAME);
```
lovisaberggren, line 35: Will this collect violations for each content type, but only one adoption per delete? I think we should move the adoption collection into the loop, or move the violation out to only count one violation per delete

wtrocki: Done. I wish we could differentiate between adoption in unit tests.

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
yelizhenden-mdb: Yes, we can simplify. The code does not do any check for other status codes, so it can be simplified to handle 204 status code only. But if you are saying we should look for all 2xx responses and check if they are empty, this is another thing, we can note it down to the spec

wtrocki: Done

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,37 @@
...
+    if (successResponse.content) {
+      for (const contentType of Object.keys(successResponse.content)) {
+        // Check if the response has a schema property
+        if (successResponse.content[contentType] && successResponse.content[contentType].schema) {
+          return collectAndReturnViolation(path, RULE_NAME, ERROR_MESSAGE);
+        }
+      }
+    }
+    collectAdoption(path, RULE_NAME);
+  }
```
lovisaberggren, line 36:
```suggestion
      for (const contentType of Object.keys(successResponse.content)) {
        // Check if the response has a schema property
        if (successResponse.content[contentType] && successResponse.content[contentType].schema) {
          return collectAndReturnViolation(path, RULE_NAME, ERROR_MESSAGE);
        }
        collectAdoption(path, RULE_NAME);
      }
    }
  }
```

wtrocki: @lovisaberggren I'm using this PR to check how easy is to write trivial rule. The collection error patterns seems to be the l problem, as it led to multiple comments on this PR.  I will look to define example high level API tomorrow and get your approval.

Unsure if we want to call adoption method multiple times in for loop etc.

lovisaberggren: > Unsure if we want to call adoption method multiple times in for loop etc.

Since we are collecting violation in the loop, we should also collect adoption. So for each content type, if it has a schema -> violation, if it doesn't, adoption. Otherwise we might have one endpoint with 3 violations for 3 content types, but only one adoption for a valid endpoint with three valid content types. If we only collect one adoption the violations will weigh more in the metrics. Let me know if that makes sense

lovisaberggren: Though since we are collecting exception only once, perhaps we should collect a violation only if one of the responses are violating, and not one for each response. It's about having the three types weigh equally

wtrocki: > It's about having the three types weigh equally

Exactly. The current state does return error/success once.
I have feeling that you want us to make change to return violations 3 and 3 sucesses here.
While I can make this change I would like us to get it documented first and agreed widely to make sure that all other rules follow the same approach.

Thank you for bearing with me trying things :)

wtrocki: CC @yelizhenden-mdb Based on your understanding for this rule should we return 1violation (with multiple errors) or multiple violations?

lovisaberggren: No strong opinion, it can be either 1 or 3, but not mixed as it is now. Based on out discussion on Slack it seems like we're leaning towards always trying to return 1,1,1, I'm good with that, so just make sure this returns one violation if at least one of the responses have a schema, instead of for all

lovisaberggren: My bad, I read it wrong, but it should be correct since it will return on the first violation found. Sorry for the confusion! I'll close the thread

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,61 @@
...
+    return collectAndReturnViolation(path, RULE_NAME, errors);
+  }
+
+  collectAdoption(path, RULE_NAME);
+};
+
+/**
+ * Check if the operation has validation issues
+ * @param {object} input - The  object to vefify
+ * @return {Array<string>|undefined} - The content types that have a schema
```
wtrocki, line 38: [JS dev standards] really important in JS to be very specific about how we going to handle errors. 
Spectral API is not giving option for array and strings.
If we assume we always return array we can validate and throw issues for the users putting invalid arguments into the method etc.

We can use undefined or base on size of the array in function above if needed.

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,61 @@
...
+            errors.push({
+              message: `Error found for ${contentType}: ${ERROR_MESSAGE}`,
+            });
+          }
+        }
+        return errors.length > 0 ? errors : undefined;
+      }
+    }
+  } catch (e) {
+    return ['Internal Rule Error without reporting violation' + e];
```
wtrocki, line 57: We can get different error when rule missbehaves. That gives us:

- Ability to understand the issue instead of `    AggregateError: Error running Nimma` 
- Protect production failure to break entire validation. Users will be able to add exception until we fix problem in single rule but other rules will still work.

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
wtrocki: I actually rubberducked myself. Changing to return single errors array.

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -0,0 +1,61 @@
...
+ * @param {object} input - The delete operation object
+ * @param {object} _ - Unused
+ * @param {object} context - The context object containing the path
+ */
+export default (input, _, { path }) => {
+  // 1. Filter out not relevant use cases that should not lead to adoption.
+  const deleteOp = input;
+  if (!deleteOp.responses || deleteOp.responses.length === 0) {
+    return;
+  }
```
wtrocki, line 18: Filter use cases might be avoided if we provide proper and very targeted given values. 
@yelizhenden-mdb Open Question:
- Do we want to handle failures in each validator for `deleteOp.responses` is not present?
- If not should we have dedicated rule that checks that every delete operation has responses

Otherwise we will have gap in validation/reporting.

yelizhenden-mdb: In my opinion, "Delete method should return an empty response" makes these assumptions:
- We are looking for delete methods that define responses
- We are looking for delete method that define a 2xx response

Given that and given we can define `ExtensionProperty` for `@ApiResponse` annotation, I would even suggest that we can define exceptions for the 2xx responses at @ApiResponse level, which do not have empty body (in our case, checking for schema property)

Something like `given: $.paths[*].delete.responses[?(@property.match(/^(2\d{2})$/))]` would work

With this `given`, it would get you all 2xx responses for delete methods, and you would not need to capture `no responses` or `no 2xx responses` cases

yelizhenden-mdb: Answer for second question: 
We can check 
- the presence of 204 No Content in the relevant rule for guideline (`Empty response should be 204 No Content`) 
- the presence of 404 Not Found in the relevant rule for guideline (`The Delete method should succeed if and only if a resource was present and was successfully deleted. If the resource did not exist, the method should send a NOT_FOUND error`)

We don't need to check the presence of responses in this rule

wtrocki: I will choose middle ground. Still rely on given $.paths[*].delete but look for exceptions inside the rule.

tools/spectral/ipa/rulesets/functions/deleteMethodResponseShouldNotHaveSchema.js
wtrocki: Moved to helper for reuse.

tools/spectral/ipa/__tests__/deleteMethodResponseShouldNotHaveSchema.test.js
```diff
@@ -0,0 +1,98 @@
...
+  },
+  {
+    name: 'valid with exception',
+    document: {
+      paths: {
+        '/resource/{id}': {
+          delete: {
+            responses: {
+              204: {
+                'application/vnd.atlas.2023-01-01+json': {
```
yelizhenden-mdb, line 82: ❤️

yelizhenden-mdb (approved):
LGTM! Great work with big impact on formalizing the rule implementation

PR #485, opened 2025-03-06
CLOUDP-271994: IPA-106: Validate Create methods accepts no query params

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-271994

Verify that the parameters (inline or referenced schemas) of POST operation does not have "in":"query"

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/spectral/ipa/__tests__/createMethodShouldNotHaveQueryParameters.test.js
+++ b/tools/spectral/ipa/__tests__/createMethodShouldNotHaveQueryParameters.test.js
@@ -1,0 +1,135 @@
+import testRule from './__helpers__/testRule';
+import { DiagnosticSeverity } from '@stoplight/types';
+
+const componentSchemas = {
+  schemas: {
+    Schema: {
+      type: 'object',
+    },
+  },
+};
+
+testRule('xgen-IPA-106-create-method-should-not-have-query-parameters', [
+  {
+    name: 'valid methods',
+    document: {
+      components: componentSchemas,
+      paths: {
+        '/resource': {
+          post: {
+            parameters: [
+              {
+                name: 'header-param',
+                in: 'header',
+                schema: { type: 'string' },
+              },
+              {
+                name: 'resource-id',
+                in: 'path',
+                schema: {
+                  $ref: '#/components/schemas/Schema',
+                },
+              },
+            ],
+          },
+        },
+        '/resource2': {
+          post: {
+            parameters: [],
+          },
+        },
+      },
+    },
+    errors: [],
+  },
+  {
+    name: 'invalid methods',
+    document: {
+      components: componentSchemas,
+      paths: {
+        '/resource': {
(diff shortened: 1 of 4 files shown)


tools/spectral/ipa/__tests__/createMethodShouldNotHaveQueryParameters.test.js
```diff
@@ -0,0 +1,135 @@
...
+  },
+  {
+    name: 'invalid methods',
+    document: {
+      components: componentSchemas,
+      paths: {
+        '/resource': {
+          post: {
+            parameters: [
+              {
```
wtrocki, line 53: How you going to handle:

```
         {
            "$ref": "#/components/parameters/envelope"
          },
```

those are quite common across our current API.

wtrocki: I recommend generating tests cases based on the existing API cases thus making self documenting what we would see as error, exception and success.

yelizhenden-mdb: The rule runs on the resolved schemas, so it should capture it, but let me ensure that they are captured. Thanks for bringing that up!

wtrocki: Sure! Here is example of one of the API that will not conform to the created method:
```
      "post": {
        "description": "Creates one project. Projects group clusters into logical collections that support an application environment, workload, or both. Each project can have its own users, teams, security, tags, and alert settings. To use this resource, the requesting API Key must have the Read Write role.",
        "operationId": "createProject",
        "parameters": [
          {
            "$ref": "#/components/parameters/envelope"
          },
          {
            "$ref": "#/components/parameters/pretty"
          },
```

yelizhenden-mdb: I included this case to the test cases. Thank you!

tools/spectral/ipa/__tests__/createMethodShouldNotHaveQueryParameters.test.js
```diff
@@ -0,0 +1,146 @@
...
+        '/resource2': {
+          post: {
+            parameters: [
+              {
+                name: 'header-param',
+                in: 'header',
+                schema: { type: 'string' },
+              },
+              {
+                $ref: '#/components/parameters/QueryParam',
```
wtrocki, line 89: It appears that your example is missing definition of #/components/parameters/QueryParam.
Assuming QueryParam from name might not be valid use case and we need to ensure that referenced parameter is in fact query parameter. Let me know if you want me to generate example for you and correct this unit test.

wtrocki (approved):

PR #492, opened 2025-03-07
CLOUDP-304053: IPA-106:Create - The resource must be the request body

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-304053

Check if the request body content of the `Create` method and the response content of the `Get` method refer to the same resource. 

**Important Note:** The request body schemas can be inline or referenced. For both, check the field equivalence. (Don't check for readOnly/writeOnly properties)

**Example**: The create and get methods for the /groups/{groupId}/clusters resource follow this rule

**Decision 1:** If a resource provides a Create method but not a Get method (`post /resource` and `get /resource/{id}` should exist at the same time), ignore it.

**Decision 2:** Check for the same versions of `create.requestBody` and `get.responses`. If there is no equivalent version in the get response, ignore it.

## Checklist

### Changes to Spectral


wtrocki:
Any way we can putthose somewhere in the repository:
Decision 1: If a resource provides a Create method but not a Get method (post /resource and get /resource/{id} should exist at the same time), ignore it.

Decision 2: Check for the same versions of create.requestBody and get.responses. If there is no equivalent version in the get response, ignore it.

Could be manual notes in readme or part of the ipa docs we generate.

tools/spectral/ipa/rulesets/functions/createMethodRequestBodyIsGetResponse.js
```diff
@@ -0,0 +1,67 @@
...
+  postMethodRequestContentPerMediaType,
+  getMethodResponseContentPerMediaType
+) {
+  const errors = [];
+
+  if (
+    !isEqual(
+      omitDeep(postMethodRequestContentPerMediaType.schema, 'readOnly', 'writeOnly'),
+      omitDeep(getMethodResponseContentPerMediaType.schema, 'readOnly', 'writeOnly')
+    )
```
wtrocki, line 58: We need to be aware of circular dependencies: https://github.com/lodash/lodash/issues/5764

wtrocki (approved):
LGTM with assumption that it handles complexities of our current schemas.
Could be good to use actual chunks of our openapi schema in tests.

- `oneOf` with returned parent objects
- Circular references

.github/workflows/spectral-lint.yml
```diff
@@ -29,6 +29,13 @@ jobs:
           sparse-checkout: |
             openapi/
             tools/spectral
+      - name: Setup Node
```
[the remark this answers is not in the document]
yelizhenden-mdb: It could not find the new libraries I included for the rule, so added these steps

tools/spectral/ipa/rulesets/functions/createMethodRequestBodyIsGetResponse.js
yelizhenden-mdb: This will compare the resolved schemas. So I am expecting it will be all value comparison but not reference comparisons

wtrocki: Yes. That was my point. I think we have >10 circular references in the models and that have been breaking postman for us: https://github.com/mongodb/openapi/blob/main/tools/postman/scripts/transform-for-api.sh#L49

Good to ensure that all current openapi violations are valid.

yelizhenden-mdb: Okay, let me add extra unit tests for this case. Do you have any example in hand, I can imitate it as unit test?

tools/spectral/ipa/rulesets/functions/utils/resourceEvaluation.js
```diff
@@ -150,3 +150,53 @@ function removePrefix(path) {
...
+ */
+export function getResponseOfGetMethodByMediaType(mediaType, pathForResourceCollection, oas) {
+  const resourcePathItems = getResourcePathItems(pathForResourceCollection, oas.paths);
+  const resourcePaths = Object.keys(resourcePathItems);
+  if (resourcePaths.length === 1) {
+    return null;
+  }
+
+  const singleResourcePath = resourcePaths.find(
+    (path) => !isCustomMethodIdentifier(path) && path !== pathForResourceCollection
```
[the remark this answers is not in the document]
yelizhenden-mdb: It would make more sense :D thanks for the catch!

package.json
```diff
@@ -29,7 +29,9 @@
         "apache-arrow": "^19.0.1",
         "dotenv": "^16.4.7",
         "eslint-plugin-jest": "^28.10.0",
+        "lodash": "^4.17.21",
```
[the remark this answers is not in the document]
yelizhenden-mdb: Fingers crossed hope we will not need custom one, but let's see :D

marcosuma:
I might have missed it somewhere else, but should also the response of the CREATE match the request of the CREATE? I guess that's probably another validation?

tools/spectral/ipa/rulesets/functions/createMethodRequestBodyIsGetResponse.js
wtrocki: Looked into something simple but.. our schema is too complex so we can have simpler version for tests.
I think all unit tests should cover oneOf and then we can replace that across

wtrocki (approved):

PR #494, opened 2025-03-07
CLOUDP-302985: Add rule xgen-IPA-104-get-method-no-request-body

lovisaberggren:
## Proposed changes

Adds rule get must not have request body. Added tests and ran the validation (no offenders currently).

_Jira ticket:_ [CLOUDP-302985](https://jira.mongodb.org/browse/CLOUDP-302985)

### DIFF
--- a/tools/spectral/ipa/__tests__/getMethodHasNoRequestBody.test.js
+++ b/tools/spectral/ipa/__tests__/getMethodHasNoRequestBody.test.js
@@ -82,19 +82,19 @@ testRule('xgen-IPA-104-get-method-no-request-body', [
     errors: [
       {
         code: 'xgen-IPA-104-get-method-no-request-body',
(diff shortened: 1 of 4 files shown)


tools/spectral/ipa/__tests__/getMethodHasNoRequestBody.test.js
```diff
@@ -0,0 +1,98 @@
...
+      },
+    ],
+  },
+  {
+    name: 'invalid with exception',
+    document: {
+      paths: {
+        '/resource/{id}': {
+          get: {
+            'x-xgen-IPA-exception': {
```
wtrocki, line 82: should exception be on request body? Not problem I genuinely asking what do you think we should do here as I have exact use case and cannot decide.

yelizhenden-mdb (approved):

PR #495, opened 2025-03-07
CLOUDP-302671: Add filter command that removes x-xgen extensions from OAS 

blva:
## Proposed changes

_Jira ticket:_ CLOUDP-302671

- Add a filter command, to be executed after foas is merged
- The command applies non-versioning filterings, clearing up any custom extensions: IPA exceptions, owner teams, and hides the endpoints for the given env if no version is provided
- It apples versioning filtering if a version is provided

### Testing
```shell
./bin/foascli filter -s ../../openapi/v2.json -o ../../openapi/v2.json --env prod
````
Result: https://github.com/mongodb/openapi/pull/495/commits/2dd3e797a2307198575e3334c7529a77c0d9f54a

## Checklist

### Changes to Spectral

### CODE REVIEW

**Comment by wtrocki** [comment]
Do you plan to enable filters for merges to master? Curious as we will need to document it for ipa local dev

File: tools/cli/internal/cli/filter/filter.go [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+// filter -s oas -o output-oas.json.
+func Builder() *cobra.Command {
+	opts := &Opts{
+		fs: afero.NewOsFs(),
+	}
+
+	cmd := &cobra.Command{
+		Use: "filter -s spec ",
+		Short: `Filter Open API specification removing hidden endpoints and extension metadata. 
+If a version is provided, versioning filters will also be applied.`,
```
andreaangiolillo, line 92: Should we define a flag with the name of the filter to apply? I am wondering a flag would help to clarify what filters will be applied to the spec

tools/cli/internal/cli/merge/merge_test.go
```diff
@@ -159,6 +159,7 @@ func TestInvalidPath_PreRun(t *testing.T) {
 		outputPath:    "foas.html",
 		externalPaths: externalPaths,
 		basePath:      "base.json",
+		format:        "json",
```
andreaangiolillo, line 162:
```suggestion
		jsonFormat:        "json",
```

tools/cli/internal/openapi/filter/operations.go
```diff
@@ -36,36 +35,9 @@ func (f *OperationsFilter) Apply() error {
 		for _, operation := range pathItem.Operations() {
 			if operation.Extensions != nil {
 				delete(operation.Extensions, "x-xgen-owner-team")
+				delete(operation.Extensions, "x-xgen-changelog")
```
andreaangiolillo, line 38: Do we want to remove also `x-xgen-IPA-exception` ?

andreaangiolillo (approved):
LGTM

tools/cli/internal/cli/filter/filter.go
blva: I thought about it, but since we don't have a use case I would rather only give the two options for now, to avoid generating several different specs with different filters applied

tools/cli/internal/openapi/filter/operations.go
blva: it's done by another filter

andreaangiolillo: I am confused 🤔 Isn't the goal here to remove all the custom extensions from an openapi spec so that integrations (AtlasCLI) can use it without worry of which extension they should parse?

blva: sorry the exception is being filtered out as well as you suggested, but in another filter file as the x-xgen-IPA-exception is in many layers of the openapi spec!

PR #497, opened 2025-03-10
CLOUDP-302671: Update v2 to filtered OAS

blva:
## Proposed changes

_Jira ticket:_CLOUDP-302671

**Context**: The OAS stored under v2.yaml and v2.json is the same used for the preview link in our repo. However, that OAS was not being filtered, containing extra extensions such as `x-xgen-hidden-env`, `x-xgen-owner-team`, IPA extensions and others. Given that this is the upper OAS in the repo, which is public, we have decided to filter out this APIx specific extensions out of them. However, we still find it very important to keep one raw, unfiltered OAS where we can see and track the full history of how extensions have changed. Furthermore, some tools will still leverage those extensions. To account for that, we opted for keeping an unfiltered raw file in a more hidden folder, to discourage external users to use it but still allow our tools and integrations to use an unfiltered file.

Changes:
- Applies and stores the filtered version of the OAS in v2.yaml/v2.json automation
- Moves the unfiltered FOAS into `.raw/v2.yaml|json`
- Points all spectral linting tools to lint on `.raw/v2.yaml|json` since both IPA and versioning linting is performed in the extensions 

Closes #[issue number]

## Checklist

### Changes to Spectral


wtrocki:
LGTM for Approach! Thank you

yelizhenden-mdb (approved):
LGTM. Thank you!

.github/scripts/split_spec.sh
```diff
@@ -6,9 +6,15 @@ foascli versions -s openapi-foas.json -o ./openapi/v2/versions.json --env "${tar
 
 echo "Running FOAS CLI split command with the following --env=${target_env:?} and -o=./openapi/v2/openapi.json"
 
+# Generate one OAS per version based on the versioning extensions
 foascli split -s openapi-foas.json --env "${target_env:?}" -o ./openapi/v2/openapi.json --format all
-mv -f "openapi-foas.json" "./openapi/v2.json"
-mv -f "openapi-foas.yaml" "./openapi/v2.yaml"
+
+# Filters out the current foas, removing all extensions that are not related with versioning
+foascli filter -s openapi-foas.json --env "${target_env:?}" -o ./openapi/v2/v2.json --format all
```
andreaangiolillo, line 13: why do we move the v2.json under the v2 folder?

.github/scripts/split_spec.sh
```diff
@@ -17,4 +23,4 @@ echo "Moving preview files to preview and private-preview folder"
 find ./openapi/v2 -type f -name "*private-preview*" -exec mv -f {} ./openapi/v2/private/ \;
 
 echo "Generate the versions.json file for private preview APIs"
-foascli versions --spec ./openapi/v2.json --stability-level PRIVATE-PREVIEW -o ./openapi/v2/private/versions.json --env "${target_env:?}"
+foascli versions --spec ./openapi/.raw/v2.json --stability-level PRIVATE-PREVIEW -o ./openapi/v2/private/versions.json --env "${target_env:?}"
```
andreaangiolillo, line 26: Could you filter out from the branded preview the v2 spec and the v2 in the .raw?

.github/scripts/split_spec.sh
blva: good catch 👁️ we shouldn't!

.github/scripts/split_spec.sh
blva: not sure I follow, but for the branded preview, I made it use the filtered version, since it would hide the hidden contents

andreaangiolillo: There is a step in the release process where we create the branded preview (branded_preview.sh), and I see that right now https://htmlpreview.github.io/?https://github.com/mongodb/openapi/blob/main/openapi/branded-preview.html we list v2.json and v2.yaml. I am wondering if with these changes we will start to see also the spec in .raw 🤔  This can also be a follow up ticket. Thanks!

blva: ohhh got it! good point, not sure and I hope not :D If we see it I'll remove in a follow-up

tools/spectral/ipa/metrics/config.js
```diff
@@ -5,7 +5,7 @@ const dirname = path.dirname(fileURLToPath(import.meta.url));
 const rootDir = path.resolve(dirname, '../../../../');
 
 const config = {
-  defaultOasFilePath: path.join(rootDir, 'openapi', 'v2.json'),
+  defaultOasFilePath: path.join(rootDir, 'openapi', '.raw', 'v2.json'),
```
wtrocki, line 8: ❤️

wtrocki (approved):

PR #503, opened 2025-03-11
task: Adding golang precommit hooks

wtrocki:
## Proposed changes

Adding specific golang precommit hooks to help with streamlining PR creation.
Changes are only run when go file change is detected in changeset.


wtrocki:
FYI @andreaangiolillo @blva

tools/hooks/golang-precommit.sh
```diff
@@ -0,0 +1,26 @@
...
+# Run golangci-lint
+echo "Running golangci-lint..."
+make lint
+
+# Run unit tests for Go files
+echo "Running Go unit tests..."
+make unit-test
+
+# Return to the original directory
+cd -
```
andreaangiolillo, line 23: can we use pushd popd here please

tools/hooks/golang-precommit.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Formatting Go files..."
+make fmt
+
+# Run golangci-lint
+echo "Running golangci-lint..."
+make lint
+
+# Run unit tests for Go files
+echo "Running Go unit tests..."
+make unit-test
```
andreaangiolillo, line 20: Since we have most of this in the cli makefile, can we have here 
```
#!/usr/bin/env bash
pushd tools/cli
make pre-commit
```

so that we reuse the logic in the make file in the cli

tools/hooks/golang-precommit.sh
wtrocki: No problem! adding.

andreaangiolillo (approved):
LGTM, thanks for all the changes

PR #509, opened 2025-03-11
CLOUDP-305831: Fix IPA validation failed on OAS release

lovisaberggren:
## Proposed changes

Optional validations failed in OAS release due to bug in rule `xgen-IPA-104-get-method-returns-response-suffixed-object`. When resolving a schema for non-2xx responses, it's undefined and wasn't checked for before checking if it has a schema.

Also added a test that will catch this if it happens again.

See failures: https://github.com/mongodb/openapi/actions/runs/13786521109

_Jira ticket:_ [CLOUDP-305831](https://jira.mongodb.org/browse/CLOUDP-305831)

### DIFF
--- a/tools/spectral/ipa/__tests__/getMethodReturnsResponseSuffixedObject.test.js
+++ b/tools/spectral/ipa/__tests__/getMethodReturnsResponseSuffixedObject.test.js
@@ -9,6 +9,15 @@ const componentSchemas = {
     Schema: {
       type: 'object',
     },
+    BadRequest: {
+      content: {
+        'application/json': {
+          schema: {
+            type: 'string',
+          },
+        },
+      },
+    },
   },
 };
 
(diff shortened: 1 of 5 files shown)


yelizhenden-mdb (approved):
Thanks!

tools/spectral/ipa/__tests__/getMethodReturnsResponseSuffixedObject.test.js
```diff
@@ -42,6 +51,9 @@ testRule('xgen-IPA-104-get-method-returns-response-suffixed-object', [
                   },
                 },
               },
+              401: {
+                $ref: '#/components/schemas/BadRequest',
+              },
```
lovisaberggren, line 56: Example of what caused the failure, since this rule operates on the unresolved schema, this will end up as undefined, since we're trying to get the schema ref inside of /content/version. Also moved the check for 2xx response first to not evaluate these

PR #516, opened 2025-03-11
task: add ipa validation to run during CI/CD process

wtrocki:
## Proposed changes

IPA validation needs to run during PR process to verify if:

- IPA rule were not made error by mistake and covers existing API
- package bump does not break spectral ruleset

PR should be never merged with failed spectral validation as it will affect and break downstream builds.

## Testing

Run echo "Ensuring no error violations available in OpenAPI"
Ensuring no error violations available in OpenAPI

> ipa-validation
> spectral lint ./openapi/v2.yaml --ruleset=./tools/spectral/ipa/ipa-spectral.yaml

No results with a severity of 'error' found!

> spectral-validation
> spectral lint ./openapi/v2.yaml --ruleset=./tools/spectral/.spectral.yaml

No results with a severity of 'error' found!

### DIFF
--- a/.github/workflows/spectral-lint.yml
+++ b/.github/workflows/spectral-lint.yml
@@ -6,16 +6,14 @@ on:
     paths: 
       - 'tools/spectral/**'
       - 'openapi/**.yaml'
-      - '.spectral.yaml'
-      - 'ipa-spectral.yaml'
+      - 'package.json'
   push:
     branches:
       - main
     paths: 
       - 'tools/spectral/**'
       - 'openapi/**.yaml'
-      - '.spectral.yaml'
-      - 'ipa-spectral.yaml'
+      - 'package.json'
 
 jobs:
   spectral-lint:
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
         "lint-js": "npx eslint **/*.js",
         "gen-ipa-docs": "node tools/spectral/ipa/scripts/generateRulesetReadme.js",
         "ipa-validation": "spectral lint ./openapi/.raw/v2.yaml --ruleset=./tools/spectral/ipa/ipa-spectral.yaml",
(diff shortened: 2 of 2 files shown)


.github/workflows/code-health-tools.yml
```diff
@@ -65,6 +65,11 @@ jobs:
       - name: Run Jest tests
         run: |
           npm run test
+      - name: Run IPA Spectral Tests against current raw OpenAPI
+        run: |
+          echo "Ensuring no error violations available in OpenAPI"
+          npm run ipa-validation
```
wtrocki, line 71: That does not run our "classic" spectral rules. Ideally we might want to run both?

wtrocki: Done

yelizhenden-mdb: How does it differentiate if there are validation failures or some other error happening?

lovisaberggren: Are these not already running on PRs? Reference: https://github.com/mongodb/openapi/blob/main/.github/workflows/spectral-lint.yml

wtrocki: @lovisaberggren Good catch. I will adjust that job to run on javascript files. 

@yelizhenden-mdb Error message will be different. Validation failure vs Spectral failure (which is unlikely since we try catch logic at the moment).

lovisaberggren: Cool, yeah that's why it didn't run in a few cases 👍

.github/workflows/code-health-tools.yml
```diff
@@ -65,6 +65,11 @@ jobs:
       - name: Run Jest tests
         run: |
           npm run test
+      - name: Run IPA Spectral Tests against current raw OpenAPI
```
lovisaberggren, line 68: Is this new one still needed as we have the spectral-lint workflow? Perhaps I missed something here

wtrocki: Reverted.

PR #530, opened 2025-03-12
DOP-5448: Integrate Bump GH action to generate spec pages

rayangler:
## Proposed changes

_Jira ticket:_ [DOP-5448](https://jira.mongodb.org/browse/DOP-5448)

The main goal of this PR is to follow Bump.sh's recommendation for a GH action that deploys changes to an OpenAPI spec to Bump, to be uploaded to the DOP team's Bump account.

## Checklist

### Changes to Spectral

## Further comments

Some implementation details to note:

* Each spec is separated into its own document on Bump under the same hub.
* I've tested this [out on a fork](https://github.com/rayangler/mdb-openapi), in case you're curious about uploads.
* I have a scoped Bump API token that must be added as a secret to allow the GH action to work properly. I can reach out on Slack about getting this added!
* For Atlas Admin API v2, we're trying to follow the current setup where each of its resource versions are separate versions (Bump calls these branches). This makes the assumption that the latest resource version will always be at the end of the `versions.json` array. Right now, the `generateSpecMapping` script is mostly for the Atlas Admin API, but hopefully the array/script can be extended to handle any future OpenAPI specs that may be added to this repo.

### DIFF
--- a/.github/scripts/generateSpecMapping.js
+++ b/.github/scripts/generateSpecMapping.js
@@ -5,6 +5,7 @@ import { fileURLToPath } from 'node:url';
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
+// Add new specs to be deployed to Bump here
 const SPEC_MAPPING = [
   {
     doc: 'f929d2d7-27d7-4591-b22f-6c2e543a7874',
@@ -13,9 +14,6 @@ const SPEC_MAPPING = [
   },
 ];
 
-/**
- * Handles the resource versions for Atlas Admin API v2
- */
 function handleAdminAPIv2() {
   const docId = '2accf4b8-a543-426c-94c3-794ae26b68be';
   const directory = 'openapi/v2';
--- a/.github/scripts/generateSpecMapping.js
+++ b/.github/scripts/generateSpecMapping.js
@@ -1,0 +1,48 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const ATLAS_ADMIN_API_V2_DOC = 'atlas-admin-api-v2';
+
+const SPEC_MAPPING = [
+  {
+    doc: 'atlas-admin-api-v1',
+    file: 'openapi/v1-deprecated/v1.json',
+    branch: 'main',
+  },
+  // Need to programmatically handle resource versions separately
+  {
+    doc: ATLAS_ADMIN_API_V2_DOC,
+    file: 'openapi/v2.json',
+    branch: 'latest',
+  },
+];
+
+/**
+ * Handles the resource versions for Atlas Admin API v2
+ */
+function handleResourceVersions() {
+  const directory = 'openapi/v2';
+  const filePath = path.join(__dirname, `../../${directory}/versions.json`);
+  const versions = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+
+  for (const version of versions) {
+    const openapiFilename = `openapi-${version}.json`;
+    const openapiFilePath = path.join(path.dirname(filePath), openapiFilename);
+
+    if (!fs.existsSync(openapiFilePath)) {
+      continue;
+    }
+
+    SPEC_MAPPING.push({
+      doc: ATLAS_ADMIN_API_V2_DOC,
+      file: `${directory}/${openapiFilename}`,
+      branch: version,
+    });
+  }
+}
+
+handleResourceVersions();
+console.log(JSON.stringify(SPEC_MAPPING));
--- a/.github/scripts/generateSpecMapping.js
+++ b/.github/scripts/generateSpecMapping.js
@@ -4,31 +4,29 @@ import { fileURLToPath } from 'node:url';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
-const ATLAS_ADMIN_API_V2_DOC = 'atlas-admin-api-v2';
 
 const SPEC_MAPPING = [
   {
-    doc: 'atlas-admin-api-v1',
+    doc: 'f929d2d7-27d7-4591-b22f-6c2e543a7874',
     file: 'openapi/v1-deprecated/v1.json',
     branch: 'main',
-  },
-  // Need to programmatically handle resource versions separately
-  {
-    doc: ATLAS_ADMIN_API_V2_DOC,
-    file: 'openapi/v2.json',
-    branch: 'latest',
   },
 ];
 
 /**
  * Handles the resource versions for Atlas Admin API v2
  */
-function handleResourceVersions() {
+function handleAdminAPIv2() {
+  const docId = '2accf4b8-a543-426c-94c3-794ae26b68be';
   const directory = 'openapi/v2';
   const filePath = path.join(__dirname, `../../${directory}/versions.json`);
   const versions = JSON.parse(fs.readFileSync(filePath, 'utf8'));
 
-  for (const version of versions) {
+  if (!versions || !Array.isArray(versions)) {
+    return;
+  }
+
+  for (const [index, version] of versions.entries()) {
     const openapiFilename = `openapi-${version}.json`;
     const openapiFilePath = path.join(path.dirname(filePath), openapiFilename);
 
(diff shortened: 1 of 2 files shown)


.github/scripts/generateSpecMapping.js
```diff
@@ -0,0 +1,55 @@
...
+const SPEC_MAPPING = [
+  {
+    doc: 'f929d2d7-27d7-4591-b22f-6c2e543a7874',
+    file: 'openapi/v1-deprecated/v1.json',
+    branch: 'main',
+  },
+];
+
+function handleAdminAPIv2() {
+  const docId = '2accf4b8-a543-426c-94c3-794ae26b68be';
```
andreaangiolillo, line 18: can we pass this ID as env variable? you can add another secret or var to the repo

.github/workflows/generate-bump-pages.yml
```diff
@@ -0,0 +1,73 @@
...
+    name: Deploy API documentation on Bump.sh
+    strategy:
+      matrix: 
+        spec-mapping: ${{ fromJson(needs.create-matrix.outputs.matrix) }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Deploy API documentation
+        uses: bump-sh/github-action@v1
```
andreaangiolillo, line 48: FYI: We ping to a gitsha when the third party gh action is not from github
```suggestion
        uses: bump-sh/github-action@690c81156715f37cb72d006e5cbb81fbd9b45365
```

.github/workflows/generate-bump-pages.yml
```diff
@@ -0,0 +1,73 @@
...
+    name: Create API preview on Bump.sh
+    strategy:
+      matrix: 
+        spec-mapping: ${{ fromJSON(needs.create-matrix.outputs.matrix) }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Create API preview
+        uses: bump-sh/github-action@v1
```
andreaangiolillo, line 67:
```suggestion
        uses: bump-sh/github-action@690c81156715f37cb72d006e5cbb81fbd9b45365
```

.github/scripts/generateSpecMapping.js
```diff
@@ -0,0 +1,55 @@
...
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+// Add new specs to be deployed to Bump here
+const SPEC_MAPPING = [
+  {
+    doc: 'f929d2d7-27d7-4591-b22f-6c2e543a7874',
```
andreaangiolillo, line 11: can we pass this ID as env variable

.github/workflows/generate-bump-pages.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/checkout@v4
+      - name: Deploy API documentation
+        uses: bump-sh/github-action@v1
+        with:
+          doc: ${{matrix.spec-mapping.doc}}
+          token: ${{secrets.BUMP_TOKEN}}
+          file: ${{matrix.spec-mapping.file}}
+          branch: ${{matrix.spec-mapping.branch}}
+  
+  api-preview:
```
andreaangiolillo, line 55: what is the experience to access the preview spec on bump once the job is completed? I am wondering if we should add a step here to add a comment to the PR with the link to the spec

.github/scripts/generateSpecMapping.js
```diff
@@ -0,0 +1,55 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+// Add new specs to be deployed to Bump here
+const SPEC_MAPPING = [
```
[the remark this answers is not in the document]
rayangler: I updated the comment to be more explicit, but let me know if you need that written out elsewhere or if you need a better explanation!

Right now, it'll create a separate job for each major spec version of the Atlas Admin API, and for each resource version. Docs currently supports 1-2 more additional public OpenAPI specs, which means we may add them to this repo, leading to more jobs, but I wasn't sure if this repo is intended for **all** OpenAPI specs, or if the preference is to only have this repo be for Atlas Admin API exclusively.

If the number of jobs becomes a concern, what we could potentially do is explore creating a custom action that can accept the entire array as a single input and bulk deploy/preview everything instead of relying on Bump's action that only accepts 1 at a time.

.github/workflows/generate-bump-pages.yml
rayangler: Right now, each matrix job will result in a separate preview link in the GH action logs ([example](https://github.com/rayangler/mdb-openapi/actions/runs/13598372933/job/38020034353#step:3:11)). Would you like me to file a separate ticket to investigate the best way to comment on the PR (probably without spamming a comment for each one) or is this sufficient for now?

.github/workflows/generate-bump-pages.yml
```diff
@@ -0,0 +1,76 @@
...
+    branches:
+      - main
+    paths:
+      - 'openapi/**.json'
+
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'openapi/**.json'
```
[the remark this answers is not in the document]
wtrocki: I do not think this is what we want for publishing our API. We do not use PR format to update openapi etc.

.github/workflows/generate-bump-pages.yml
```diff
@@ -0,0 +1,76 @@
+name: Check & deploy API documentation
+
+on:
+  workflow_dispatch: # Allow manual trigger in case of quick fix
+  push:
+    branches:
+      - main
```
wtrocki, line 7: you intend to update on production only? No dev branch?

wtrocki (approved):
LGTM for approach. 
Let's remove ability to trigger this action for now before we merge PR.

.github/workflows/generate-bump-pages.yml
wtrocki: > Right now, each matrix job will result in a separate preview link in the GH action logs

SGTM. Feel free to introduce it in separate PR and spec out how to get access to preview urls.

.github/workflows/generate-bump-pages.yml
wtrocki: Answered offline.

.github/scripts/generateSpecMapping.js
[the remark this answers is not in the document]
wtrocki: SGTM. It is not about how many jobs but how many minutes are used. 
I see that in current scenario this new job will use less than 10 minutes of run time monthly.

.github/scripts/generateSpecMapping.js
```diff
@@ -0,0 +1,57 @@
...
+        file,
+        branch: 'latest',
+      });
+    }
+  }
+}
+
+handleAdminAPIv2();
+// Output to GH action
+console.log(JSON.stringify(SPEC_MAPPING));
```
caesarbell, line 57: Nice call, this way we can see the mapping in GH Actions.

caesarbell (approved):
LGTM

andreaangiolillo (approved):
LGTM. Thanks for all the changes

.github/scripts/generateSpecMapping.js
```diff
@@ -0,0 +1,57 @@
...
+// for Atlas Admin API v2 will lead to a new entry in the array with its own Bump branch.
+const SPEC_MAPPING = [
+  {
+    doc: process.env.ATLAS_ADMIN_V1_DOC_ID,
+    file: 'openapi/v1-deprecated/v1.json',
+    branch: 'main',
+  },
+];
+
+function handleAdminAPIv2() {
```
blva, line 19: nit: can we please have unit tests?

blva: (ok to add in a follow-up PR)

.github/scripts/generateSpecMapping.js
```diff
@@ -0,0 +1,57 @@
...
+  const docId = process.env.ATLAS_ADMIN_V2_DOC_ID;
+  const directory = 'openapi/v2';
+  const filePath = path.join(__dirname, `../../${directory}/versions.json`);
+  const versions = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+
+  if (!versions || !Array.isArray(versions)) {
+    return;
```
blva, line 26: nit: add logs for troubleshooting in the future, if file changes and we fail to parse

.github/scripts/generateSpecMapping.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  for (const [index, version] of versions.entries()) {
+    const openapiFilename = `openapi-${version}.json`;
+    const openapiFilePath = path.join(path.dirname(filePath), openapiFilename);
+
+    if (!fs.existsSync(openapiFilePath)) {
```
blva, line 33: nit: add logs if we find a version but not a file

.github/scripts/generateSpecMapping.js
rayangler: Good call. Added

.github/scripts/generateSpecMapping.js
rayangler: Okay, I can file a follow-up DOP ticket. Is there an ideal location within the repo to add this test? I didn't see a test directory for existing action scripts, but maybe I missed it 😅

blva: good point! I am not sure, in our case, we have scripts but they call some of the tools we built under the tools folder, which is where we have the most logic, if you think you'll build more on top of this script you can look into https://github.com/mongodb/openapi/tree/main/tools/spectral/ipa and create a tool folder for `docs` and have tests there, otherwise we can look into GH best practices on how to add coverage for the scripts!

PR #532, opened 2025-03-13
Fix: Refine Create method descriptions with the resource collection applicability

yelizhenden-mdb:
## Proposed changes

Includes "This rule applies only to POST requests targeting resource collection URIs." sentence to the IPA-106:Create rules

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/spectral/ipa/rulesets/IPA-106.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-106.yaml
@@ -10,7 +10,7 @@ functions:
 
 rules:
   xgen-IPA-106-create-method-request-body-is-request-suffixed-object:
-    description: |
+    description: >-
       The Create method request should be a Request suffixed object. http://go/ipa/106
       This rule applies only to POST requests targeting resource collection URIs.
     message: '{{error}} http://go/ipa/106'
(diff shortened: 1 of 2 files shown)


lovisaberggren (approved):

tools/spectral/ipa/rulesets/IPA-106.yaml
```diff
@@ -10,24 +10,29 @@ functions:
 
 rules:
   xgen-IPA-106-create-method-request-body-is-request-suffixed-object:
-    description: 'The Create method request should be a Request suffixed object. http://go/ipa/106'
+    description: |
```
wtrocki, line 13: Do not use | as it will break the table in markdown
```suggestion
    description: >-
```

wtrocki:
@yelizhenden-mdb @lovisaberggren YAML has multiple formats for multilines. Since our generator uses table format any form of line breaks or markdown like content will break table. To prevent from that we can use `>-` 

Alternatively we can move from table format to chapter format per rule. IMHO that could be much better as each rule can be linked directly in failure mesasge. WDYT?

wtrocki (approved):
LGTM. Happy to apply suggestions across the implementation

wtrocki:
Merged as I would like to get this one refined as well.

PR #535, opened 2025-03-13
CLOUDP-304938: Adds rule xgen-IPA-105-list-method-response-is-get-method-response

lovisaberggren:
## Proposed changes

Adds rule `xgen-IPA-105-list-method-response-is-get-method-response` which checks that the paginated List method results reference the same schema as the resource Get method. Also handles version matching, which affected the existing rule `xgen-IPA-106-create-method-request-body-is-get-method-response`.

New rule `xgen-IPA-105-list-method-response-is-get-method-response` has 2 violations
Existing rule `xgen-IPA-106-create-method-request-body-is-get-method-response` has one new violation
I'll do follow-up to add exceptions to both above

Added tests for the rule and unit tests for the functions used to get the responses based on a version.

_Jira ticket:_ [CLOUDP-304938](https://jira.mongodb.org/browse/CLOUDP-304938)


tools/spectral/ipa/rulesets/functions/utils/resourceEvaluation.js
```diff
@@ -150,51 +150,3 @@ function removePrefix(path) {
   }
   return path;
 }
-
```
lovisaberggren, line 153: Moved to methodUtils

tools/spectral/ipa/rulesets/functions/utils/methodUtils.js
```diff
@@ -37,8 +40,116 @@ export function getAllSuccessfulResponseSchemas(operationObject) {
...
+  if (!getMethodObject.responses) {
+    return null;
+  }
+
+  const response = getMethodObject.responses['200'];
+  if (!response || !response.content) {
+    return null;
+  }
+
+  const versions = Object.keys(response.content);
```
lovisaberggren, line 136: Added version matching here used in both `getResponseOfListMethodByMediaType` and `getResponseOfGetMethodByMediaType`

tools/spectral/ipa/rulesets/IPA-105.yaml
```diff
@@ -29,3 +30,18 @@ rules:
     then:
       field: '@key'
       function: 'eachResourceHasListMethod'
+  xgen-IPA-105-list-method-response-is-get-method-response:
+    description: >-
+      The response body of the List method should consist of the same resource object returned by the Get method. Validation checks that the List method results items reference the same schema as the Get method response.<br/>
+      - Validation applies to List methods for resource collections only<br/>
```
yelizhenden-mdb, line 36: As far as I understand:
- Validation makes schemaName level comparison only
- Validation also does not make inline schema comparison

Are these assumptions correct? If so,
List-Get relationship is strict, so I am okay with expecting exactly the same schema.
Do we assume Get response always has schema reference, and should we consider inline schemas?

lovisaberggren: > Validation makes schemaName level comparison only

Yes, it checks that it refers to the exact same schema

> Validation also does not make inline schema comparison

The paginated schema can be inline, but the resource schema itself is expected to be a schema ref, if the get doesn't have a ref, there will be a violation. I can add a test to cover this to make sure

tools/spectral/ipa/rulesets/functions/listMethodResponseIsGetMethodResponse.js
```diff
@@ -0,0 +1,109 @@
...
+    responseCode !== '200' ||
+    !mediaType.endsWith('json') ||
+    !isResourceCollectionIdentifier(resourcePath) ||
+    (isResourceCollectionIdentifier(resourcePath) && isSingletonResource(getResourcePathItems(resourcePath, oas.paths)))
+  ) {
+    return;
+  }
+
+  // Ignore if the List method does not have a response schema
+  const listMethodResponse = getResponseOfListMethodByMediaType(mediaType, resourcePath, oas);
```
yelizhenden-mdb, line 42: Why do we call this method? Isn't the rule iterating list method response schemas already?

lovisaberggren: Good catch, we can just get the response, I'll update 👍

tools/spectral/ipa/rulesets/IPA-105.yaml
```diff
@@ -29,3 +30,18 @@ rules:
     then:
       field: '@key'
       function: 'eachResourceHasListMethod'
+  xgen-IPA-105-list-method-response-is-get-method-response:
+    description: >-
+      The response body of the List method should consist of the same resource object returned by the Get method. Validation checks that the List method results items reference the same schema as the Get method response.<br/>
```
wtrocki, line 35: That is based on assumption that "list operations" return pagginated objects - with results, total etc.
I seen that one before. Would we have validation and helper to check for that assumption  each rule can base on it?
```suggestion
      The response body of the List method should consist of the same resource object returned by the Get method. Validation checks that the List method response contains items property with reference the same schema as the Get method response.<br/>
```

tools/spectral/ipa/rulesets/IPA-105.yaml
```diff
@@ -29,3 +30,18 @@ rules:
...
+      - Validation applies to List methods for resource collections only<br/>
+      - Validation applies to json response content only<br/>
```
wtrocki, line 37: No longer needed. You need to merge master if that were needed.
```suggestion
      - Validation applies to json response content only
```

lovisaberggren: Thanks!

tools/spectral/ipa/rulesets/IPA-105.yaml
```diff
@@ -29,3 +30,18 @@ rules:
...
+      - Validation ignores responses without schema and non-paginated responses<br/>
+      - Validation ignores resources without a Get method<br/>
+      - Paths with `x-xgen-IPA-exception` for this rule are excluded from validation<br/>
+      http://go/ipa/105
+    message: '{{error}} http://go/ipa/105'
```
wtrocki, line 42: use ipa-spectral instead: https://github.com/mongodb/openapi/pull/537

It will provide link to all your implementation details. IPA link is also part of that documentation.

tools/spectral/ipa/rulesets/functions/listMethodResponseIsGetMethodResponse.js
```diff
@@ -0,0 +1,115 @@
...
+  const listSchemaRef = getSchemaRef(listMethodResponse.schema);
+  if (!listSchemaRef) {
+    resolvedListSchema = listMethodResponse.schema;
+  } else {
+    const listSchemaName = getSchemaNameFromRef(listSchemaRef);
+    resolvedListSchema = resolveObject(oas, ['components', 'schemas', listSchemaName]);
+  }
+
+  // Ignore if the List method response is not found or not paginated
+  if (!resolvedListSchema || !schemaIsPaginated(resolvedListSchema)) {
```
wtrocki, line 59: Unrelated to the PR but helper implementation might need additional checks as it now based on weak assumption that property named results means paggination. It should be covered in the rule description etc.

 ```return fields.includes('properties') && Object.keys(schema['properties']).includes('results');```

tools/spectral/ipa/rulesets/IPA-105.yaml
lovisaberggren: Done 👍

tools/spectral/ipa/rulesets/functions/listMethodResponseIsGetMethodResponse.js
wtrocki: IMHO we should at least check if results type is array
 ```return fields.includes('properties') && Object.keys(schema['properties']).includes('results');```

lovisaberggren: Yeah, I'll add a note in the description, and also add a check that the results is an array

tools/spectral/ipa/rulesets/functions/listMethodResponseIsGetMethodResponse.js
```diff
@@ -0,0 +1,115 @@
...
+    responseCode !== '200' ||
+    !mediaType.endsWith('json') ||
+    !isResourceCollectionIdentifier(resourcePath) ||
+    (isResourceCollectionIdentifier(resourcePath) && isSingletonResource(getResourcePathItems(resourcePath, oas.paths)))
+  ) {
+    return;
+  }
+
+  // Ignore if the List method does not have a response schema
+  const listMethodResponse = oas.paths[resourcePath].get.responses['200'].content[mediaType];
```
yelizhenden-mdb, line 37:
```suggestion
  const listMethodResponse = resolveObject(oas, path);
```

yelizhenden-mdb:
```suggestion
  const listMethodResponse = resolveObject(documentInventory.resolved, path);
```

tools/spectral/ipa/rulesets/functions/listMethodResponseIsGetMethodResponse.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+function checkViolationsAndReturnErrors(path, listMethodResultItems, getMethodResponseContent) {
+  try {
+    // Error if the Get method does not have a schema
+    if (!getMethodResponseContent.schema) {
+      return [
+        {
+          path,
+          message: `Could not validate that the List method returns the same resource object as the Get method. The Get method does not have a schema.`,
```
yelizhenden-mdb, line 89: Q: Should we return the same error for the createMethodRequest comparison?

lovisaberggren: Sure, that makes sense 👍

tools/spectral/ipa/rulesets/IPA-105.yaml
wtrocki: @lovisaberggren Yes. I'm aware. My intention for this comment was to add note to the yaml/docs.

tools/spectral/ipa/rulesets/functions/listMethodResponseIsGetMethodResponse.js
wtrocki: Awesome. Second comment can be follow up jira as you haven't introduced that helper in this PR

tools/spectral/ipa/rulesets/IPA-105.yaml
```diff
@@ -29,3 +30,23 @@ rules:
     then:
       field: '@key'
       function: 'eachResourceHasListMethod'
+  xgen-IPA-105-list-method-response-is-get-method-response:
+    description: >-
+      The response body of the List method should consist of the same resource object returned by the Get method. http://go/ipa/105
```
wtrocki, line 35: Link to IPA is no longer needed as it is already part of the docs and in failure we do reference to the docs in open source repo
```suggestion
      The response body of the List method should consist of the same resource object returned by the Get method. 
```

wtrocki: Meta comment/ Not this PR/Project:

In retrospect we could use public shortener to link to this repo to be kind for non MongoDB use cases. 
Technically you could make blog post covering new mongodb validations and then go links will be the only blocker. All the info will be availabile upstream.
This can be addressed once we get our plans around IPA publication done.

wtrocki (approved):
LGTM

All comments are nits.

yelizhenden-mdb (approved):
LGTM! Thanks!

PR #543, opened 2025-03-14
CLOUDP-305862: replace urls to use public url shortener and direct link to rule

wtrocki:
## Proposed changes

Use public shortener instead of go links to link to the rule specification (generated Readme). 
Created as single PR across all files to avoid confusion in future rule development


tools/spectral/ipa/__tests__/IPA105exceptionExtensionFormat.test.js
```diff
@@ -51,25 +51,25 @@ testRule('xgen-IPA-005-exception-extension-format', [
     errors: [
       {
         code: 'xgen-IPA-005-exception-extension-format',
-        message: 'IPA exceptions must have a valid rule name and a reason. http://go/ipa-spectral#IPA-102',
+        message: 'IPA exceptions must have a valid rule name and a reason.',
```
lovisaberggren, line 54:
```suggestion
        message: 'IPA exceptions must have a valid rule name and a reason. http://go/ipa-spectral#IPA-005',
```

tools/spectral/ipa/__tests__/__helpers__/testRule.js
```diff
@@ -17,7 +17,7 @@ export default (ruleName, tests) => {
 
         errors.forEach((error, index) => {
           expect(error.code).toEqual(testCase.errors[index].code);
-          expect(error.message).toEqual(testCase.errors[index].message);
+          expect(error.message).toMatch(testCase.errors[index].message);
```
lovisaberggren, line 20: Q: Why change to match?

wtrocki: you can use regexp if needed. So basically it is equals when you need equals and regular expression when it is needed. Idea is that we do not need to put lenghtly links in the tests

wtrocki: e.g in future if we change url shortener no one will need to spent +1 hour adjusting test covering the error url

tools/spectral/ipa/__tests__/IPA105exceptionExtensionFormat.test.js
lovisaberggren: Ah I see now it's because of the matcher, nvm!

lovisaberggren (approved):
Thanks!

PR #553, opened 2025-03-18
CLOUDP-306347: isSingleResourceIdentifier for delete operations

wtrocki:
## Proposed changes

Refined DELETE operations validation rules (IPA-108) to only apply to single resource endpoints with path parameters. This ensures proper REST semantics where DELETE should operate on specific resources rather than collections.

### Changes include:
- Updated documentation to clarify rule application scope
- Added test cases to verify correct behavior for both resource and collection endpoints
- Added path parameter detection using the `isSingleResourceIdentifier` helper

Reused helper function created by @yelizhenden-mdb. Thank you for refinement and implementation!

### DIFF
--- a/tools/spectral/ipa/rulesets/functions/IPA108DeleteMethod204Response.js
+++ b/tools/spectral/ipa/rulesets/functions/IPA108DeleteMethod204Response.js
@@ -4,6 +4,7 @@ import {
   collectException,
   handleInternalError,
 } from './utils/collectionUtils.js';
+import { isSingleResourceIdentifier } from './utils/resourceEvaluation.js';
 import { hasException } from './utils/exceptions.js';
 
 const RULE_NAME = 'xgen-IPA-108-delete-method-return-204-response';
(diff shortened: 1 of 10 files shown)


tools/spectral/ipa/rulesets/functions/IPA108DeleteMethod204Response.js
```diff
@@ -17,6 +18,17 @@ const ERROR_MESSAGE = 'DELETE method should return 204 No Content status code.';
...
  */
 export default (input, _, { path }) => {
+  // Check if the path is for a single resource (e.g., has path parameter)
+  // Extract the path from context.path which is an array
+  const pathString = path[1]; // Assuming path is ['paths', '/resource/{id}', 'delete']
+  if (!isSingleResourceIdentifier(pathString)) {
+    return;
+  }
+
+  if (!input || typeof input !== 'object') {
```
yelizhenden-mdb, line 28: Q: Is this check necessary? We assume that this function will only be called with a defined (non-undefined/null) input

tools/spectral/ipa/rulesets/functions/IPA108DeleteMethodResponseShouldNotHaveSchema.js
```diff
@@ -16,6 +17,23 @@ const ERROR_MESSAGE = 'DELETE method should return an empty response. The respon
  * @param {object} context - The context object containing the path
  */
 export default (input, _, { path }) => {
+  // Since this rule is applied to the response object (204 response),
+  // we need to extract the path from the context.path array differently
+  // Assuming path is like ['paths', '/resource/{id}', 'delete', 'responses', '204']
+
+  if (path.length < 3) {
```
yelizhenden-mdb, line 24: Q: Why do we need this check?

wtrocki: There is no sense to check if paths length is not meet. I will remove those use case and follow approach in other rules.

yelizhenden-mdb (approved):
LGTM! Thanks!

PR #568, opened 2025-03-19
CLOUDP-307110: Combine separated put and patch IPA rules

lovisaberggren:
## Proposed changes

We have two rules for Update Method that are separated per PUT and PATCH, this PR combined them into one rule. Will update affected exceptions in follow-up.

_Jira ticket:_ [CLOUDP-307110](https://jira.mongodb.org/browse/CLOUDP-307110)


yelizhenden-mdb (approved):
LGTM

PR #588, opened 2025-03-20
CLOUDP-306580: avoid mixed refs

wtrocki:
## Proposed changes

 
CLOUDP-306580

During adding exceptions I noticed additional check is needed. 
We not only should check for mixed integers but also for integers and refs. 
This way we get much cleaner error messages etc.


yelizhenden-mdb (approved):
LGTM

PR #589, opened 2025-03-21
IPA-112: Field names must use camelCase

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-271999

Validates schema field names using the Spectral `casing` function

Drive-by fix: `IPA-112: Avoid project field names` will also check requestBody and response schemas similar to `Field names must use camelCase`

## Checklist

### Changes to Spectral


tools/spectral/ipa/__tests__/IPA112FieldNamesAreCamelCase.test.js
```diff
@@ -0,0 +1,106 @@
...
+          SchemaName: {
+            properties: {
+              UserId: { type: 'string' },
+              user_id: { type: 'string' },
+              'user-id': { type: 'string' },
+            },
+          },
+        },
+      },
+    },
```
lovisaberggren, line 37: Q: Does it also catch nested schemas, and arrays etc.?

yelizhenden-mdb: - It should catch nested schemas. I can add a unit test to be sure.
- What do you mean by catching arrays in this context? Can you give an example?

lovisaberggren: Thanks!

For arrays:
```
          ArraySchema: {
            type: 'array',
            items: {
              properties: {
                INVALID_FIELD: {},
              },
            },
          },
```

tools/spectral/ipa/rulesets/IPA-112.yaml
```diff
@@ -29,3 +30,17 @@ rules:
...
+      Schema field names should be in camelCase format.
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+        - Searches through all schemas in the API definition
+        - Identifies property names that are not in camelCase format
+        - Reports any instances where these field names are not in camelCase format
+    message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-112-field-names-are-camel-case'
+    severity: warn
+    given: '$.components.schemas..properties[*]~'
```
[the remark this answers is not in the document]
yelizhenden-mdb: Great question! We should clarify what we mean by "field." To me, a field is a property of a resource schema, and resource schemas are typically defined under `components.schemas`. However, when I think about what I’d consider a resource schema, I’d primarily refer to response or requestBody content schemas as well.

So, are we specifically talking about inline schemas within those? Or, do we also count inline schemas used for query parameters? I’ve noticed that query parameters often use inline schemas, but they fall into a gray area—they're not path parameters, (the IPA for path params does not apply to them) and I’m not sure if their properties should be considered as fields

tools/spectral/ipa/__tests__/IPA112FieldNamesAreCamelCase.test.js
yelizhenden-mdb: You mean the `properties` in `items`. Sure, let me add unit tests covering both and see how it works

tools/spectral/ipa/rulesets/IPA-112.yaml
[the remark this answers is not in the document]
lovisaberggren: Yeah I think it could be valuable to catch fields in requests and responses as well, I'm working on the description IPA rule now and included them like so:
```
'$.paths[*][get,put,post,delete,options,head,patch,trace]..content..properties[*]'
```

In terms of query parameters, it's a good point. We don't talk a lot about them in the IPAs, but I'm thinking they should also be camelCase. Perhaps we can bring a new guideline for this. I think we can leave them out of this PR at least

lovisaberggren (approved):

PR #592, opened 2025-03-21
CLOUDP-304967: Add IPA rule for description format

lovisaberggren:
## Proposed changes

Adds IPA rules `xgen-IPA-117-description-starts-with-uppercase` and `xgen-IPA-117-description-ends-with-period`.

`xgen-IPA-117-description-ends-with-period` will ignore descriptions ending with `|` as some descriptions have inline tables.

~20 violations for components and rule `xgen-IPA-117-description-ends-with-period` that were not caught with the legacy rule. Will go and fix these after merge (exceptions shouldn't be needed).

_Jira ticket:_ [CLOUDP-304967](https://jira.mongodb.org/browse/CLOUDP-304967)


tools/spectral/ipa/rulesets/functions/IPA117DescriptionUpperCasePeriod.js
```diff
@@ -0,0 +1,46 @@
...
+  const errors = checkViolationsAndReturnErrors(input['description'], path);
+  if (errors.length !== 0) {
+    return collectAndReturnViolation(path, RULE_NAME, errors);
+  }
+  collectAdoption(path, RULE_NAME);
+};
+
+function checkViolationsAndReturnErrors(description, path) {
+  const upperCaseStart = new RegExp(`^[A-Z]`);
+  const periodEnd = new RegExp(`[.]$`);
```
gssbzn, line 32: you need to consider the old case https://github.com/mongodb/openapi/blob/3915e0a1a921c00e2ccdb30a37e8b0b8a2120f5f/tools/spectral/.spectral.yaml#L56C10-L56C27 a lot of descriptions end with a markdown table `|`

lovisaberggren: Ah thanks for pointing it out. We have another guideline to not allow inline tables, so perhaps in this case if it ends with a table we should ignore rather than say it's adopted

yelizhenden-mdb: There's another IPA guideline stating (and there will be another validation rule for) `Descriptions should not include inline tables, as this may not work well with all tools, particularly generated client code`. Should we consider descriptions ending with Markdown tables valid?

lovisaberggren: Yeah that's what I was thinking, so for this rule instead of saying `.` and `|` is valid, we can ignore cases where it ends with `|` and not collect adoption

yelizhenden-mdb: I saw your comment after updating the page, sorry :D I am okay with ignoring them.

lovisaberggren: NP! Will update 👍

tools/spectral/ipa/rulesets/IPA-117.yaml
```diff
@@ -30,3 +32,52 @@ rules:
       - '$.components.parameters[*]'
     then:
       function: 'IPA117HasDescription'
+  xgen-IPA-117-description-starts-with-uppercase:
```
yelizhenden-mdb, line 35: Q: Can't we merge them into same rule as you did before, and giving inline table marks as functionOptions

lovisaberggren: Was thinking about it initially though I didn't want to skip validating that the first letter is uppercase because it ends with `|`, so I split them into two to have clearer logic in the validator. Though we could add this as a function option for the `description-ends-with-period`, wdyt?

wtrocki: Good context. Based on your explanation current state in PR SGTM

yelizhenden-mdb (approved):
LGTM

PR #595, opened 2025-03-24
Add overrides to IPA spectral

lovisaberggren:
Added some overrides which catch violations where exceptions are not possible, or violations to be addressed later.


yelizhenden-mdb (approved):
Thank you!

PR #600, opened 2025-03-25
CLOUDP-308303: size of postman file

wtrocki:
## Proposed changes

_Jira ticket:_CLOUDP-308303

POC for Postman to use js engine to enable follow up transformations to reduce size of the file.


wtrocki:
Rejected idea. Closing

PR #601, opened 2025-03-25
CLOUDP-271998: IPA-110: Pagination (Validate Paginated prefix and results array field)

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-271998 and CLOUDP-304955

This PR validates
- if 200 OK responses of List methods refer to a schema with "`Paginated`" prefix. (Related guideline amendment PR: https://github.com/10gen/ipa/pull/64)
- if 200 OK responses of List methods have a schema with results (array) fields

**Clarification**: The associated guideline the `xgen-IPA-110-collections-response-define-results-array `rule validating is `The response for collections must define an array of results containing the paginated resource`

This guideline is shared with `xgen-IPA-105-list-method-response-is-get-method-response` rule. The rule is already checking if the results array containing the paginated resource when the results array exists. Different than that, `xgen-IPA-110-collections-response-define-results-array` rule only checks if the List method response have a schema with results (array) fields.

## Follow-up
- Having `links` field following `ApiListView` implementation will be validated in scope of another guideline `The response for collections should define an array of links field, providing the user with links to the next and previous pages`
- Having `totalCount` field is not required as it is mentioned in the `The response for collections may provide an integer totalCount field, providing the user with the total number of resources available in the backing collection. This total may be an estimate but the API should explicitly document that`
- For all the other Pagination guidelines, refer to [IPA-110: Pagination](https://docs.devprod.prod.corp.mongodb.com/ipa/110) to be implemented

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/spectral/ipa/__tests__/IPA110CollectionsUsePaginatedSchema.test.js
+++ b/tools/spectral/ipa/__tests__/IPA110CollectionsUsePaginatedSchema.test.js
@@ -33,7 +33,7 @@ const componentSchemas = {
       },
     },
   },
-  NonPaginatedList: {
+  PaginatedList: {
     type: 'object',
     properties: {
       count: {
@@ -92,7 +92,7 @@ testRule('xgen-IPA-110-collections-use-paginated-schema', [
     errors: [],
   },
   {
-    name: 'valid schemas without Paginated prefix but with correct structure',
+    name: 'invalid schemas without Paginated prefix but with correct structure',
     document: {
       paths: {
         '/resources': {
@@ -118,7 +118,15 @@ testRule('xgen-IPA-110-collections-use-paginated-schema', [
         schemas: componentSchemas,
       },
     },
-    errors: [],
+    errors: [
+      {
+        code: 'xgen-IPA-110-collections-use-paginated-schema',
+        message:
+          'List methods response must reference a paginated response schema. The response should reference a schema with "Paginated" prefix',
+        path: ['paths', '/resources', 'get', 'responses', '200', 'content', 'application/json'],
+        severity: DiagnosticSeverity.Warning,
+      },
+    ],
   },
   {
     name: 'invalid list methods without correct structure',
@@ -129,11 +137,15 @@ testRule('xgen-IPA-110-collections-use-paginated-schema', [
             responses: {
               200: {
                 content: {
-                  'application/json': {
+                  'application/vnd.atlas.2024-08-05+json': {
                     schema: {
-                      $ref: '#/components/schemas/NonPaginatedList',
+                      $ref: '#/components/schemas/PaginatedList',
                     },
                   },
+                  'application/vnd.atlas.2024-01-01+json': {
+                    schema: {},
+                  },
+                  'application/vnd.atlas.2024-03-03+json': {},
                 },
               },
             },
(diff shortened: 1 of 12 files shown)


tools/spectral/ipa/__tests__/IPA110CollectionsUsePaginatedSchema.test.js
```diff
@@ -0,0 +1,336 @@
...
+      },
+      components: {
+        schemas: componentSchemas,
+      },
+    },
+    errors: [
+      {
+        code: 'xgen-IPA-110-collections-use-paginated-schema',
+        message:
+          'List methods must use a paginated response schema. The response should reference a schema with either a name starting with "Paginated" or contain both "totalCount" (integer) and "results" (array) fields.',
```
lovisaberggren, line 154:
```suggestion
          'List methods must use a paginated response schema. The response should reference a schema with a name starting with "Paginated" and contain both "totalCount" (integer) and "results" (array) fields.',
```
On the error message, I think we should have the proper guidance that it's not enough to just prefix the name with `Paginated`

tools/spectral/ipa/__tests__/IPA110CollectionsUsePaginatedSchema.test.js
```diff
@@ -0,0 +1,336 @@
...
+      components: {
+        schemas: componentSchemas,
+      },
+    },
+    errors: [
+      {
+        code: 'xgen-IPA-110-collections-use-paginated-schema',
+        message:
+          'The schema is defined inline and must reference a predefined schema with a name starting with "Paginated".',
+        path: ['paths', '/resources', 'get', 'responses', '200', 'content', 'application/json'],
```
lovisaberggren, line 299:
```suggestion
          'The schema is defined inline and must reference a predefined paginated schema.',
        path: ['paths', '/resources', 'get', 'responses', '200', 'content', 'application/json'],
```
[Nit]

tools/spectral/ipa/__tests__/IPA110CollectionsUsePaginatedSchema.test.js
```diff
@@ -0,0 +1,336 @@
...
+        severity: DiagnosticSeverity.Warning,
+      },
+    ],
+  },
+  {
+    name: 'valid schema with exception',
```
lovisaberggren, line 305:
```suggestion
    name: 'invalid schema with exception',
```

tools/spectral/ipa/rulesets/IPA-110.yaml
```diff
@@ -0,0 +1,22 @@
...
+rules:
+  xgen-IPA-110-collections-use-paginated-schema:
+    description: |
+      APIs that return collections of resources must use a paginated response schema.
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+        - Only applies to List methods (GET operations that return collections of resources)
+        - Checks that the 200 response schema references a schema with a name starting with either a name starting with "Paginated" or contain both "totalCount" (integer) and "results" (array) fields
+        - If the referenced schema name has "Paginated" prefix, the rule assumes that the schema is paginated and does not check for "totalCount" and "results" fields
```
lovisaberggren, line 16: Should we assume this? Can we do the proper check for totalCount and results regardless of the name?

tools/spectral/ipa/rulesets/functions/IPA110CollectionsUsePaginatedSchema.js
```diff
@@ -0,0 +1,87 @@
...
+  ) {
+    return;
+  }
+
+  // Ignore if the List method does not have a response schema
+  const listMethodResponse = resolveObject(oas, path);
+
+  if (!listMethodResponse || !listMethodResponse.schema) {
+    return;
+  }
```
lovisaberggren, line 39: Wondering if we should error here, since the List should have a response?

tools/spectral/ipa/__tests__/IPA110CollectionsUsePaginatedSchema.test.js
yelizhenden-mdb: I was thinking if we force people to use Paginated prefix, but why not? It brings consistency across the naming. There are some cases extending `ApiListView`, but the name does not have `Paginated` prefix.

I think I can also update the IPA guideline to be more clear on how we decide if it is paginated with the same words here. Wdyt?

tools/spectral/ipa/rulesets/functions/IPA110CollectionsUsePaginatedSchema.js
yelizhenden-mdb: Yes it makes sense, it is proposing that List method should have a response and how the response should look like

tools/spectral/ipa/rulesets/IPA-110.yaml
yelizhenden-mdb: Done 👍

tools/spectral/ipa/rulesets/functions/IPA110CollectionsUsePaginatedSchema.js
yelizhenden-mdb: Done 👍

wtrocki:
`Paginated` prefix is weak requirement IMHO as we do not validate that in any way. 
It is derrived from the class name. Do we see classes with `Paginated` that do not conform to th definition of page.
If not let's remove `Paginated` prefix.

wtrocki:
> Assumption: If schema name starts with Paginated, it will be assumed that it extends ApiListView, and the referenced schema will not be validated

That information should be present in the openapi already. Using name as assumption might lead to false negatives.

tools/spectral/ipa/__tests__/IPA110CollectionsUsePaginatedSchema.test.js
lovisaberggren: Related to wojciechs comment, I have no strong opinion but we should either require Paginated prefix or not, but we should always check for results and total count regardless of the schema name

lovisaberggren (approved):

wtrocki (changes requested):
Suggesting to use reliance on the base class or check for format of the fields instead of the name.

wtrocki (approved):
Disscussed offline. Yeliz to make changes based on new IPA proposed rule. 
Unblocking

PR #605, opened 2025-03-25
CLOUDP-304968: IPA rule xgen-IPA-117-description-should-not-use-inline-tables

lovisaberggren:
## Proposed changes

Adds rule `xgen-IPA-117-description-should-not-use-inline-tables` which checks with regex if the description contains characters denoting a markdown table, specifically the separator between the table header and the rows (`|---|`, `|:---|`, `|---:|` or `|:---:|`).

It's still allowed to write for example `Value is: this | that`.

Checked the violations (currently 33) to ensure no false positives.

_Jira ticket:_ [CLOUDP-304968](https://jira.mongodb.org/browse/CLOUDP-304968)

### DIFF
--- a/tools/spectral/ipa/rulesets/functions/IPA117DescriptionShouldNotUseTables.js
+++ b/tools/spectral/ipa/rulesets/functions/IPA117DescriptionShouldNotUseTables.js
@@ -1,0 +1,42 @@
+import { hasException } from './utils/exceptions.js';
+import {
+  collectAdoption,
+  collectAndReturnViolation,
+  collectException,
+  handleInternalError,
+} from './utils/collectionUtils.js';
+
+const RULE_NAME = 'xgen-IPA-117-description-should-not-use-inline-tables';
+const ERROR_MESSAGE =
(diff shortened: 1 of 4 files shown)


tools/spectral/ipa/rulesets/functions/IPA117DescriptionShouldNotUseTables.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const errors = checkViolationsAndReturnErrors(input['description'], path);
+  if (errors.length !== 0) {
+    return collectAndReturnViolation(path, RULE_NAME, errors);
+  }
+  collectAdoption(path, RULE_NAME);
+};
+
+function checkViolationsAndReturnErrors(description, path) {
+  const tablePattern = new RegExp(`[|]:?-+:?[|]`);
```
wtrocki, line 32: https://stackoverflow.com/questions/9837935/regex-for-markdown-table-syntax

wtrocki (approved):

PR #628, opened 2025-03-27
CLOUDP-306573: IPA rule - plaintext repsonse has example

lovisaberggren:
## Proposed changes

New rule `xgen-IPA-117-plaintext-response-must-have-example` checks that 2xx responses (non-json and non-yaml) have examples in the object or in the schema.

Also updated the IPA validation that runs on the PRs to use the spectral CLI instead of the action, since the action runs on a downloaded spec it can't show the annotations in the PR as a result. The spectral CLI output is a bit more useful as it shows all warnings/errors and the component that is violating.

Before: 

Screenshot 2025-03-27 at 17 45 07

After:

Screenshot 2025-03-27 at 17 44 58

_Jira ticket:_ [CLOUDP-306573](https://jira.mongodb.org/browse/CLOUDP-306573)


lovisaberggren:
IPA failure related to `xgen-IPA-113-singleton-should-have-update-method` which has a fix on MMS side

.github/workflows/spectral-lint.yml
```diff
@@ -44,8 +44,5 @@ jobs:
           file_glob: openapi/.raw/v2.yaml
           spectral_ruleset: tools/spectral/.spectral.yaml #If updated, need to update in MMS too.
       - name: IPA validation action
-        uses: stoplightio/spectral-action@2ad0b9302e32a77c1caccf474a9b2191a8060d83
-        with:
-          file_glob: v2.yaml
-          spectral_ruleset: tools/spectral/ipa/ipa-spectral.yaml
+        run: npx spectral lint v2.yaml --ruleset=./tools/spectral/ipa/ipa-spectral.yaml
```
yelizhenden-mdb, line 47: ❤️

tools/spectral/ipa/rulesets/IPA-117.yaml
```diff
@@ -157,3 +158,20 @@ rules:
...
+      For APIs that respond with plain text, for example CSV, API producers must provide an example. Some tools are not able to generate examples for such responses
+
+      ##### Implementation details
+        - The rule only applies to 2xx responses
+        - The rule ignores JSON and YAML responses (passed as `allowedTypes`)
+        - The rule checks for the presence of the example property as a sibling to the `schema` property, or inside the `schema` object
+    message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-117-plaintext-response-must-have-example'
+    severity: warn
+    given:
+      - '$.paths[*][get,put,post,delete,options,head,patch,trace].responses[*].content'
```
yelizhenden-mdb, line 172: Q: I know this comes a long way but.. Do we need to specify them? Do you remember what is the reason behind specifying HTTP methods?

wtrocki (approved):

tools/spectral/ipa/rulesets/IPA-117.yaml
lovisaberggren: Yes, if we don't specify them this will also target other things like extension properties as well. For this line specifically the risk is pretty low to have an extension with responses and content, but just as a safe-guard

PR #637, opened 2025-03-31
IPA-114: Errors (fix for exception level)

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-304960

Fixes:
- Exceptions will only be collected at the response level.
- There might be multiple errors for one response, assuming there might be several versions of responses (CLOUDP-276969)
- Add error if referenced schema is not valid
(`openapi-api-registry.json` refers to schemas in `openapi-mms.json`, which cannot be resolved in custom Spectral rule functions, therefore they will appear as errors)
- For investigating external schema reference resolution further, CLOUDP-309694 is created

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/spectral/ipa/rulesets/IPA-114.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-114.yaml
@@ -14,6 +14,7 @@ rules:
       This rule checks that all 4xx and 5xx error responses reference the ApiError schema.
     message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-114-error-responses-refer-to-api-error'
     severity: warn
+    resolved: false
     given: '$.paths[*][*].responses[?(@property.match(/^[45]\d\d$/))]'
     then:
       function: 'IPA114ErrorResponsesReferToApiError'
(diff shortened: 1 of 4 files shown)


tools/spectral/ipa/rulesets/IPA-114.yaml
```diff
@@ -14,6 +14,7 @@ rules:
       This rule checks that all 4xx and 5xx error responses reference the ApiError schema.
     message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-114-error-responses-refer-to-api-error'
     severity: warn
+    resolved: false
```
lovisaberggren, line 17: FYI this is probably why the external spec references don't work, since we're passing the unresolved spec

One option to consider could be to run the validation on the resolved spec, but when evaluating the object, use the passed `unresolvedDocument` to see the schema ref name

yelizhenden-mdb: Discussed offline. For now, we will proceed with the current approach, and for the API Registry violations, I will note down the limitation in the exception reason, and refer to the investigation ticket

lovisaberggren (approved):

PR #638, opened 2025-03-31
CLOUDP-306573: Dont require gzip to have example

lovisaberggren:
## Proposed changes

Updates IPA rule `xgen-IPA-117-plaintext-response-must-have-example` to ignore file types. Added a test to cover case.

_Jira ticket:_ [CLOUDP-306573](https://jira.mongodb.org/browse/CLOUDP-306573)

### DIFF
--- a/tools/spectral/ipa/rulesets/IPA-117.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-117.yaml
@@ -166,7 +166,7 @@ rules:
 
       ##### Implementation details
         - The rule only applies to 2xx responses
-        - The rule ignores JSON and YAML responses (passed as `allowedTypes`)
+        - The rule ignores JSON, YAML and GZIP responses (passed as `allowedTypes`)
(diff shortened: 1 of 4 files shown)


tools/spectral/ipa/rulesets/IPA-117.yaml
```diff
@@ -176,7 +176,7 @@ rules:
       field: '@key'
       function: 'IPA117PlaintextResponseMustHaveExample'
       functionOptions:
-        allowedTypes: ['json', 'yaml']
+        allowedTypes: ['json', 'yaml', 'gzip']
```
yelizhenden-mdb, line 179: Q: Instead of adding it as `gzip`, how about checking if the schema format is `binary` and ignoring them accordingly? That way, we wouldn't need to list all future possible binary media types

lovisaberggren: Sure, that's a good idea. It'll ignore other filetypes if there would be new ones in the future

yelizhenden-mdb (approved):
LGTM. Thank you!

PR #640, opened 2025-03-31
CLOUDP-304963: IPA-114: Errors (check for 404 codes for parameterized endpoints)

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-304963

```
  xgen-IPA-114-parameterized-paths-have-404-not-found:
    description: |
      Paths with parameters must define 404 responses.

      ##### Implementation details
      This rule checks that all endpoints with path parameters (identified by '{param}' 
      in the path) include a 404 response to handle the case when the requested resource
      is not found.
```

## Checklist

### Changes to Spectral

### DIFF
--- a/tools/spectral/ipa/rulesets/IPA-114.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-114.yaml
@@ -5,6 +5,7 @@ functions:
   - IPA114ErrorResponsesReferToApiError
   - IPA114ApiErrorHasBadRequestDetail
   - IPA114AuthenticatedEndpointsHaveAuthErrors
+  - IPA114ParameterizedPathsHave404NotFound
 
 rules:
   xgen-IPA-114-error-responses-refer-to-api-error:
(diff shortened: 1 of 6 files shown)


lovisaberggren (approved):

tools/spectral/ipa/rulesets/IPA-114.yaml
```diff
@@ -45,3 +46,16 @@ rules:
...
+    description: |
+      Paths with parameters must define 404 responses.
+
+      ##### Implementation details
+      This rule checks that all endpoints with path parameters (identified by '{param}' 
+      in the path) include a 404 response to handle the case when the requested resource
+      is not found.
+    message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-114-parameterized-paths-have-not-found'
+    severity: warn
+    given: '$.paths[*][get,put,post,delete,options,head,patch,trace]'
```
wtrocki, line 59: [Outside PR] Delete has it's own rule. If if introduce generic rule (which I think makes sense) we should remove dedicated 404 rule for delete. WDYT?

wtrocki: Separately - what is the purpose of mentioning all of the methods here. 
I guess you wanted to exclude exceptions? If yes the same filtering can be done in the code.

wtrocki (approved):
LGTM with follow up recommendation.

tools/spectral/ipa/rulesets/IPA-114.yaml
yelizhenden-mdb: Yes, I wanted to exclude exceptions and all the stuff other than HTTP verbs. I agree that it can be handled in the code, but it is widely used approach at the moment, so I followed it

We can remove the dedicated 404 rule for delete, as you said it will be covered by this one. I will share the PR for removing it in a separate PR, if that is okay

PR #673, opened 2025-04-08
CLOUDP-311375: Update `foascli split` command to support `upcoming` apis

andreaangiolillo:
## Proposed changes
CLOUDP-311375

This PR updates `foascli split` to add support for `upcoming` api version.

### DIFF
--- a/tools/cli/internal/apiversion/version.go
+++ b/tools/cli/internal/apiversion/version.go
@@ -65,6 +65,10 @@ func (v *APIVersion) newVersion(version string, date time.Time) {
 	if IsPreviewStabilityLevel(version) {
 		v.stabilityVersion = PreviewStabilityLevel
 	}
+
+	if IsUpcomingStabilityLevel(version) {
+		v.stabilityVersion = UpcomingStabilityLevel
+	}
 }
 
 // WithVersion sets the version on the APIVersion.
@@ -166,7 +170,7 @@ func (v *APIVersion) StabilityLevel() string {
 }
 
 func (v *APIVersion) ExactMatchOnly() bool {
-	return v.IsPreview()
+	return v.IsPreview() || v.IsUpcoming()
 }
 
 func (v *APIVersion) IsPreview() bool {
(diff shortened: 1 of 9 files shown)


tools/cli/internal/apiversion/version.go
```diff
@@ -166,7 +170,7 @@ func (v *APIVersion) StabilityLevel() string {
 }
 
 func (v *APIVersion) ExactMatchOnly() bool {
-	return v.IsPreview()
+	return v.IsPreview() || v.IsUpcoming()
```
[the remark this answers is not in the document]
blva: how do we plan to release upcoming OASes, I think that if we go with exact match, same approach will happen where we only have the endpoints for upcoming in the OAS, if that's the goal, all good! otherwise we need to do date matching with the date

tools/cli/internal/cli/split/split_test.go
```diff
@@ -78,6 +78,26 @@ func TestSplitPrivatePreviewRun(t *testing.T) {
...
 }
 
+func TestSplitUpcomingRun(t *testing.T) {
+	t.Parallel()
+	fs := afero.NewMemMapFs()
+	opts := &Opts{
+		basePath:   "../../../test/data/openapi_with_upcoming.json",
+		outputPath: "foas.yaml",
+		fs:         fs,
+		env:        "dev",
```
blva, line 88: the-absolute-super-nit: set format to yaml  so that you only generate a yaml as "all" is the default and we'll generate an unused file

tools/cli/internal/apiversion/version.go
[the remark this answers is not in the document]
andreaangiolillo: yes, we want to have only upcoming APIs in the upcoming OAS 👍  Thanks for double-checking

tools/cli/internal/cli/split/split_test.go
andreaangiolillo: I will update it

blva (approved):
LGTM!

tools/cli/internal/cli/split/split_test.go
blva: sorry i approved can be in your next PR

PR #688, opened 2025-04-16
Add Aliases to IPA Validation Rulesets

lovisaberggren:
Adds aliases to IPA Spectral rulesets to minimize duplication and increase reusability of similar `given` statements. Note: Aliases cannot be inherited from other spectral files, so they must be added to each yaml


yelizhenden-mdb (approved):
LGTM! Thanks for the refactoring!

PR #731, opened 2025-05-23
CLOUDP-314000: Clarify IPA-124 maxItems rule description

lovisaberggren:
## Proposed changes

Based on feedback there was some confusion about the max allowed value for the `maxItems` property, since "default" was mentioned, it can be confused with the openapi property `default` or that this field is set to 100 by default. 

Clarified description of rule that 100 is the recommended max size of an array, and that the `maxItems` property value should not be larger than that.

_Jira ticket:_ [CLOUDP-314000](https://jira.mongodb.org/browse/CLOUDP-314000)


yelizhenden-mdb (approved):
LGTM. Thanks!

PR #748, opened 2025-06-09
CLOUDP-311382: Update `foas changelog metadata` command to support `upcoming`

andreaangiolillo:
## Proposed changes
CLOUDP-311382

This PR proposes the following changes:
- updates the `foas changelog metadata create` to consider upcoming API version when generating the metadata file
- updates the `foas changelog create` to skip upcoming api version when generating the changelog. The support will be added in CLOUDP-315486

## Next
I will test the changes once the PR is merged by running the FOAS release. I may have to create follow up PRs to fix the release flow.

### DIFF
--- a/tools/cli/internal/changelog/changelog.go
+++ b/tools/cli/internal/changelog/changelog.go
@@ -180,8 +180,8 @@ func NewEntries(basePath, revisionPath, exceptionFilePath string) ([]*Entry, err
 	}
 
 	for _, version := range changelog.RevisionMetadata.Versions {
-		// Skip preview versions
-		if apiversion.IsPreviewStabilityLevel(version) {
+		// Skip preview and upcoming versions
+		if apiversion.IsPreviewStabilityLevel(version) || apiversion.IsUpcomingStabilityLevel(version) {
 			continue
 		}
 
@@ -229,6 +229,12 @@ func NewEntriesBetweenRevisionVersions(revisionPath, exceptionFilePath string) (
 			if apiversion.IsPreviewStabilityLevel(fromVersion) || apiversion.IsPreviewStabilityLevel(toVersion) {
 				continue
 			}
+
+			// Skip upcoming preview version. It will be included in CLOUDP-315486
+			if apiversion.IsUpcomingStabilityLevel(fromVersion) || apiversion.IsUpcomingStabilityLevel(toVersion) {
+				continue
+			}
+
 			entry, err := newEntriesBetweenVersion(revisionMetadata, fromVersion, toVersion, exceptionFilePath)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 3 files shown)


tools/cli/internal/changelog/changelog.go
```diff
@@ -229,6 +229,12 @@ func NewEntriesBetweenRevisionVersions(revisionPath, exceptionFilePath string) (
 			if apiversion.IsPreviewStabilityLevel(fromVersion) || apiversion.IsPreviewStabilityLevel(toVersion) {
 				continue
 			}
+
+			// Skip upcoming preview version. It will be included in CLOUDP-315486
+			if apiversion.IsUpcomingStabilityLevel(fromVersion) || apiversion.IsUpcomingStabilityLevel(toVersion) {
```
blva, line 234: maybe we only add it once it turns into the version, no? so we would never add upcoming to the changelog?

tools/cli/internal/cli/changelog/metadata/create.go
```diff
@@ -70,7 +71,10 @@ func (o *Opts) PreRun() error {
 		}
 	}
 
+	// Validate that the API version use the correct date format YYYY-MM-DD
 	for _, version := range o.versions {
+		// Upcoming version has the format YYYY-MM-DD.upcoming, here we remove .upcoming to validate the date format.
+		version = strings.ReplaceAll(version, ".upcoming", "")
```
blva, line 77: this is where I suggested to simply try to use `apiversion.go` and create a new version, throwing an error if it's invalid, that way you don't need to maintain two different logics to parse versions

blva (approved):
non-blocking comments

tools/cli/internal/cli/changelog/metadata/create.go
andreaangiolillo: yes, good point. I updated the logic to reuse `apiversion.go`

blva (approved):
LGTM thanks for the changes

tools/cli/internal/cli/changelog/metadata/create_test.go
```diff
@@ -47,6 +47,18 @@ func TestCreateBuild_PreRun_InvalidVersion(t *testing.T) {
 	require.ErrorContains(t, opts.PreRun(), "invalid version date")
 }
 
+func TestCreateBuild_PreRun_UpcomingVersion(t *testing.T) {
+	fs := afero.NewMemMapFs()
+	opts := &Opts{
+		specRevision: "test",
+		runDate:      "2024-01-01",
+		versions:     []string{"2024-01-01.upcoming", "2025-01-01"},
```
wtrocki, line 55: Q: is there chance for us to migrate from string to object with multiple keys in the future? 
 Not sure how much involved it would be, asking curious question.

 Thinking of all code generation tools we need to invent some parser for custom date format and prefixes

PR #752, opened 2025-06-10
CLOUDP-323790: Update foas release process to support bump.sh deployment for all the envs

andreaangiolillo:
## Proposed changes
This PR updates the foas release process to support bump.sh deployment for all the envs


gssbzn (approved):

.github/scripts/generateSpecMapping.js
```diff
@@ -12,7 +12,7 @@ const SPEC_MAPPING = [
   {
     doc: process.env.ATLAS_ADMIN_V1_DOC_ID,
     file: 'openapi/v1-deprecated/v1.json',
-    branch: 'main',
+    branch: process.env.BRANCH_NAME
```
rayangler, line 15: The branch field in this case is Bump's term for a version, and non-versioned docs have a single `main` branch by default. My understanding is that we'd want to always keep this `main` since the v1 API doc is not versioned, instead of basing it off of the GH branch. 

Please correct me if I'm misunderstanding!

andreaangiolillo: you are correct. The v1 is not versioned. I was under the impression that this `main` was the branch of this repo `mongodb/openapi`. I will revert this change. Thanks!

rayangler (approved):

PR #773, opened 2025-06-16
CLOUDP-325047: add support for extension in the main element of the OAS

andreaangiolillo:
## Proposed changes
CLOUDP-325047:

Related to https://github.com/10gen/mms/pull/131979, bump recently added support for a custom extension. Here I am adding the support for extensions to the struct we use to represent an open api spec.

### DIFF
--- a/tools/cli/internal/openapi/openapi.go
+++ b/tools/cli/internal/openapi/openapi.go
@@ -34,6 +34,7 @@ type Spec struct {
 	Paths        *openapi3.Paths               `json:"paths" yaml:"paths"`
 	Components   *openapi3.Components          `json:"components,omitempty" yaml:"components,omitempty"`
 	ExternalDocs *openapi3.ExternalDocs        `json:"externalDocs,omitempty" yaml:"externalDocs,omitempty"`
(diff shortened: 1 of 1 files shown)


tools/cli/internal/openapi/openapi.go
```diff
@@ -34,6 +34,7 @@ type Spec struct {
 	Paths        *openapi3.Paths               `json:"paths" yaml:"paths"`
 	Components   *openapi3.Components          `json:"components,omitempty" yaml:"components,omitempty"`
 	ExternalDocs *openapi3.ExternalDocs        `json:"externalDocs,omitempty" yaml:"externalDocs,omitempty"`
+	Extensions   map[string]any                `json:"-" yaml:"-"`
```
gssbzn, line 37: doesn't `-` means ignored? do you really mean to ignore this field?

gssbzn (approved):
LGTM

tools/cli/internal/openapi/openapi.go
andreaangiolillo: you are right, this does not work. I am adding the missing part here https://github.com/mongodb/openapi/pull/774

PR #823, opened 2025-07-16
CLOUDP-328313: Quickfix for legacy custom methods

sphterry:
Debug for legacy custom methods case 

_Jira ticket:_ [CLOUDP-328313](https://jira.mongodb.org/browse/CLOUDP-328313)

### DIFF
--- a/tools/spectral/ipa/rulesets/functions/utils/operationIdGeneration.js
+++ b/tools/spectral/ipa/rulesets/functions/utils/operationIdGeneration.js
@@ -30,9 +30,10 @@ export function generateOperationID(method, path) {
   nouns = nouns.map((noun) => capitalize(noun));
 
   let verb = deriveActionVerb(method);
+  const camelCaseCustomMethod = method.length > verb.length;
 
   // if custom method name is multiple words, add trailing nouns to the operation ID
-  if (method.length > verb.length) {
+  if (camelCaseCustomMethod) {
     nouns.push(method.slice(verb.length));
   }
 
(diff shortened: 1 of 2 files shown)


tools/spectral/ipa/rulesets/functions/utils/operationIdGeneration.js
```diff
@@ -24,14 +24,16 @@ export function generateOperationID(method, path) {
   // legacy custom method - use end of path as custom method name
```
wtrocki, line 24: [nit] can be separate method for handling legacy operations right?

tools/spectral/ipa/rulesets/functions/utils/operationIdGeneration.js
```diff
@@ -24,6 +24,7 @@ export function generateOperationID(method, path) {
...
   if (!method) {
     method = nouns.pop();
+    resourceIdentifier = resourceIdentifier.slice(0, resourceIdentifier.lastIndexOf('/'));
```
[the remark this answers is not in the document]
wtrocki: [nit] Do we need test for empty operationIds/resourceIdentifier not in the shape we expected.

tools/spectral/ipa/rulesets/functions/utils/operationIdGeneration.js
sphterry: Yes, I could probably make a method that contains the extra logic and calls on this function. Will be something to consider when I'm implementing the verbosity override - might refactor for use in generating a shorter recommended opId

PR #824, opened 2025-07-21
DOP-5830: Update previews to use Bump

rayangler:
## Proposed changes

With the move to Bump.sh for rendering OpenAPI spec content on the MongoDB docs site, it makes sense to update the preview script to generate previews through Bump as well. Unlike Redoc, Bump doesn't have a client-side way to generate previews (that I know of at the moment), and instead relies on using [their API](https://developers.bump.sh/operation/operation-post-previews). To avoid needing to make many unnecessary calls to Bump's API, the generated HTML will now create individual temporary Bump preview links on click. A serverless Netlify function maintained by DOP will wrap the calls to Bump's API to minimize CORS issues.

Here's an example of it in action using local dev data:

https://github.com/user-attachments/assets/f7a6d287-cde3-4fc9-a3f0-c233b2749357

_Jira ticket:_ [DOP-5830](https://jira.mongodb.org/browse/DOP-5830)


andreaangiolillo (approved):
LGTM

PR #844, opened 2025-07-28
chore: upgrade oasdiff to the latest released version

andreaangiolillo:
## Proposed changes
This PR updates oasdiff to the latest release. 

## Note that oasdiff moved the repo from tufin to oasdiff so the PR contains imports renaming.


tools/cli/test/data/changelog/new-upcoming-version/output/2024-08-05_2025-09-22.upcoming.json
```diff
@@ -6,13 +6,8 @@
     "tag": "OpenAPI",
     "changes": [
       {
-        "change": "removed the media type 'application/json' for the response with the status '200'",
-        "changeCode": "response-media-type-removed",
```
blva, line 10: [q] can we check everywhere that we use these IDs? I think we might have something in breakinig changes/changelog to customize text/or breaking changes and probably need to update

blva: also exemptions for instance will stop working for the ids removed

blva (approved):

PR #861, opened 2025-08-01
fix: version check script should look at HEAD

sphterry:
## Proposed changes
Bug in version check compares main to itself - updated to look back to HEAD instead

### DIFF
--- a/.github/scripts/ipa_version_check.sh
+++ b/.github/scripts/ipa_version_check.sh
@@ -1,16 +1,12 @@
-#!/bin/bash  
-set -eou pipefail
+#!/bin/bash      
+set -eou pipefail    
+    
+current_version=$(jq -r '.version' tools/spectral/ipa/package.json)    
   
-current_version=$(jq -r '.version' tools/spectral/ipa/package.json)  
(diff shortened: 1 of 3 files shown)


.github/scripts/ipa_version_check.sh
```diff
@@ -1,16 +1,12 @@
...
-set -eou pipefail
+#!/bin/bash      
+set -eou pipefail    
+    
+current_version=$(jq -r '.version' tools/spectral/ipa/package.json)    
   
-current_version=$(jq -r '.version' tools/spectral/ipa/package.json)  
-
-previous_version=$(git show origin/"${BASE_BRANCH}":tools/spectral/ipa/package.json | jq -r '.version')  
+previous_version=$(git show HEAD~1:tools/spectral/ipa/package.json | jq -r '.version' || echo "none")
```
wtrocki, line 6: Any reason why we do not use here github defined base branch env vars?
Also you are doing sparse checkouts in actions which might affect history.

wtrocki (approved):
Fingers crossed

PR #868, opened 2025-08-05
fix(ipa): Account for abbreviations in op ID names

lovisaberggren:
## Proposed changes

This change accounts for abbreviations in operation ID names for validation, since we have some paths with names like `mongoDB` and `customDNS`. When matching/counting words we need to allow for some words to be initialized in capital letters

_Jira ticket:_ [CLOUDP-335986](https://jira.mongodb.org/browse/CLOUDP-335986)


sphterry:
LGTM!

PR #897, opened 2025-08-15
feat(ipa): new rule xgen-IPA-117-operation-summary-format

lovisaberggren:
## Proposed changes

New rule xgen-IPA-117-operation-summary-format, checks that the `summary` of an operation is in Title Case. This also covers the rules that the summary should not use CommonMark or end with period.

Will follow-up with more operation summary rules, split into one rule per PR.

_Jira ticket:_ [CLOUDP-319114](https://jira.mongodb.org/browse/CLOUDP-319114)

### DIFF
--- a/tools/spectral/ipa/rulesets/IPA-117.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-117.yaml
@@ -11,6 +11,7 @@ functions:
   - IPA117PlaintextResponseMustHaveExample
   - IPA117ObjectsMustBeWellDefined
   - IPA117ParameterHasExamplesOrSchema
+  - IPA117OperationSummaryFormat
 
 aliases:
   OperationObject:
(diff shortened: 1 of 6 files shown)


tools/spectral/ipa/rulesets/IPA-117.yaml
```diff
@@ -222,3 +223,62 @@ rules:
...
+          - 'JWKS'
+          - 'X.509'
+          - 'M2'
+          - 'M5'
+          - 'RAM'
+          - 'LTS'
+          - 'VPC'
+          - 'DN'
+          - 'CSV'
+        grammaticalWords:
```
yelizhenden-mdb, line 266: Q: We have `grammaticalWords` defined for another rule. I am curious if it is possible to reuse them somehow if they are same. Do you know if there is a way for it?

lovisaberggren: Hmm, I don't think so 🤔  Not aware of how that could work

yelizhenden-mdb: I could not find how it would work too. Looking for some global aliases, but it is not possible I guess

PR #932, opened 2025-09-01
fix(ipa): make operation ID rules error-level

lovisaberggren:
## Proposed changes

Makes the op ID rules error-level.

Some corrections/extension in the prod OAS have not yet been released. Silencing these in the meantime.

_Jira ticket:_ [CLOUDP-340284](https://jira.mongodb.org/browse/CLOUDP-340284)


tools/spectral/ipa/__tests__/metrics/data/test-ipa-spectral.yaml
```diff
@@ -1,2 +1,24 @@
...
+functions:
+  - IPA104ValidOperationID
+
+aliases:
+  GetOperationObject:
+    - '$.paths[*].get'
+
+rules:
+  xgen-IPA-104-valid-operation-id-warn:
+    description: Test rule as warning for testing purposes
```
lovisaberggren, line 15: Added this rule as warning level just for testing purposes in the metrics collection test. This spectral file is only used for the test

PR #1205, opened 2026-04-14
feat: add resources for listing specs in the registry and get a summary for a specific one

yelizhenden-mdb:
## Proposed changes

Two new MCP resources registered in the server:
- `openapi://specs` — lists all specs loaded in the registry (returns `alias`, `sourceType`, and `filePath` per spec)
- `openapi://specs/{alias}` — returns a structural overview of a single spec: `title`, `description`, `stats (paths, operations, schemas, tags)`, `latestStableVersion`, `availableVersions`, `hasPreview`, `hasUpcoming`

Version extraction reuses foascli's ExtractVersions and post-processes the result, so any new stability level foascli adds (e.g. future types) is picked up automatically.

_Jira ticket:_ CLOUDP-393142

## Checklist

### Changes to Spectral


andmatei (approved):
LGTM

tools/mcp-server/internal/resources/uri.go
```diff
@@ -0,0 +1,20 @@
+package resources
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// aliasFromURI extracts the alias from openapi://specs/{alias}.
+func aliasFromURI(uri string) (string, error) {
```
maks-m-mongo-leaf, line 11: Do we have test coverage for this code?
what would happen if "https://goodle.com/q" gets passed into this function?

tools/mcp-server/internal/resources/resources.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// makeSpecsHandler creates the handler for the openapi://specs resource.
+func makeSpecsHandler(reg *registry.Registry) mcp.ResourceHandler {
+	return func(_ context.Context, req *mcp.ReadResourceRequest) (*mcp.ReadResourceResult, error) {
+		return handleSpecs(reg, req)
+	}
+}
+
+// makeAliasHandler creates the handler for the openapi://{alias} resource template.
```
maks-m-mongo-leaf: openapi://{alias}  or openapi://specs/{alias}  ?

tools/mcp-server/internal/resources/alias_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/mongodb/openapi/tools/mcp-server/internal/registry"
+)
+
+// TestHandleAlias_Overview verifies that the spec overview contains title, stats, and version info.
+func TestHandleAlias_Overview(t *testing.T) {
+	reg := newTestRegistry(t)
+
+	result, err := handleAlias(reg, makeRequest("openapi://specs/test-api"))
+	if err != nil {
```
maks-m-mongo-leaf: Why not use `assert/require` functionality in tests?
https://pkg.go.dev/github.com/stretchr/testify/assert#NoError
https://pkg.go.dev/github.com/stretchr/testify/require#NoError

tools/mcp-server/internal/resources/resources.go
```diff
@@ -0,0 +1,52 @@
...
+		Description: "Start here. Lists all OpenAPI specifications currently loaded in the registry. " +
+			"Each entry includes the alias (used to reference the spec in all other resources and tools), " +
+			"sourceType ('file' for specs loaded from disk, 'virtual' for sliced subsets), " +
+			"and filePath (empty for virtual specs). " +
+			"Read this resource first to discover what aliases are available before using other resources or tools.",
+		MIMEType: mimeTypeJSON,
+	}, makeSpecsHandler(reg))
+
+	server.AddResourceTemplate(&mcp.ResourceTemplate{
+		URITemplate: "openapi://specs/{alias}",
```
maks-m-mongo-leaf, line 26: Does it support `openapi://specs/parent/child-alias`?

yelizhenden-mdb: No. What would `child-alias` represent in this context, slices derived from the main spec?

Either way, I don’t think keeping a parent–child relationship adds much value here, since each one already has a unique alias

tools/mcp-server/internal/resources/resources.go
yelizhenden-mdb: Fixed them, thanks!

tools/mcp-server/internal/resources/uri.go
yelizhenden-mdb: Fair concern. I changed the code to be more specific to the resource template and added test coverage. Let me know how it looks

tools/mcp-server/internal/resources/alias_test.go
yelizhenden-mdb: Thanks for flagging this! I added the testify package for test development and updated the existing tests for resources. I will open another PR for tools tests

wtrocki:
@yelizhenden-mdb Are we at the point of being able to build the MCP and test the changes?

yelizhenden-mdb:
> Are we at the point of being able to build the MCP and test the changes?


@wtrocki Yes, you can build it using the Makefile, which will produce a Go binary. You can then point your MCP client to that binary. One thing to note: make sure to use the `load` tool first to register a spec, once that’s done, the new resources will become available (first resources are introduced with this PR, so you'll need to checkout this one and generate the binary)

tools/mcp-server/internal/resources/resources.go
andmatei: I think maybe here it would be some value for virtual specs generated from slicing to keep track of the parent spec they came from instead? What do you think?

maks-m-mongo-leaf: I put it as hypothetical use case just by looking it code. I was wondering if the code would break and how if it would, if anybody would use an unexpected value.

yelizhenden-mdb: Let's assume there are one file-based spec, `atlas`, and one sliced (virtual) spec, `atlas-clusters`

IMHO, the only thing an agent needs to work with a spec is its alias. Knowing that `atlas-clusters` came from `atlas` doesn't change how the agent queries it or what it does with the result. It's metadata without a clear action behind it.

Where it could add value is in debugging context, seeing the lineage of a sliced spec. But I don't think it would add value for agent consumption

tools/mcp-server/internal/resources/uri.go
maks-m-mongo-leaf: We could achieve the following in a few ways. One is regex parsing, just to mention it. The other is url parsing. The latter is more high level and has its upsides.

If we go with the high level parsing I'd suggest to consider refactoring the duplicated `fmt.Errorf(...` line. But the code makes sense and it is more restrictive now.

tools/mcp-server/internal/resources/resources.go
maks-m-mongo-leaf: Thanks for all the context. I know a little more now.
I will resolve this comment. The other comment and the change made for that comment made this comment redundant, but at least I learnt. Thanks!

tools/mcp-server/internal/resources/uri.go
maks-m-mongo-leaf: I will resolve the comment because the last change I requested is more of a NIT.

PR #1231, opened 2026-04-16
feat(ipa): New rule for IPA 117 - requests/responses must have schema

lovisaberggren:
## Proposed changes

New warning level IPA rule for IPA 117.

Requests and responses must be well defined. I.e. if `requestBody` or `responses` are defined for an operation, the contents must be well-defined (have a schema)

The rule checks all request and response definitions, except for:
- 202 Accepted responses (it's acceptable these have no content)
- DELETE 2xx responses (same reason as above)
- Custom method 2xx responses for POST operations (custom methods are custom, and there is no requirement they return a response body)

_Jira ticket:_ [CLOUDP-393698](https://jira.mongodb.org/browse/CLOUDP-393698)

Current violations:
```
# Stable APIs missing 200 response schema:

paths./api/atlas/v2/groups/{groupId}/collStats/metrics.get.responses[200].content.application/vnd.atlas.2023-11-15+json

paths./api/atlas/v2/groups/{groupId}/maintenanceWindow.patch.responses[200].content.application/vnd.atlas.2023-01-01+json

paths./api/atlas/v2/groups/{groupId}/pushBasedLogExport.patch.responses[200].content.application/vnd.atlas.2023-01-01+json


# Sunset API missing 200 response schema:

paths./api/atlas/v2/groups/{groupId}/pushBasedLogExport.post.responses[200].content.application/vnd.atlas.2023-01-01+json
```

## Follow-up

Follow-up steps:
- Release new IPA version and run new version in upstream repo
- Check for stability as warning-level rule for ~2 weeks
- Open tickets for teams owning violations
- When confident, move the rule severity to error, release new version and update upstream repo

### DIFF
--- a/tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
+++ b/tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
@@ -48,9 +48,11 @@ testRule('xgen-IPA-117-request-response-bodies-must-be-well-defined', [
                   },
                   // Valid schema
                   'application/vnd.atlas.2023-08-05+json': {
-                    properties: {
-                      name: {
-                        type: 'string',
+                    schema: {
+                      properties: {
+                        name: {
+                          type: 'string',
+                        },
                       },
                     },
                   },
@@ -76,6 +78,9 @@ testRule('xgen-IPA-117-request-response-bodies-must-be-well-defined', [
               content: {
                 // Valid schema
                 'application/vnd.atlas.2023-08-05+json': {
+                  schema: {
+                    type: 'string',
+                  },
                   example: {
                     name: 'Test',
                   },
@@ -91,7 +96,9 @@ testRule('xgen-IPA-117-request-response-bodies-must-be-well-defined', [
                 content: {
                   // Valid schema
                   'application/vnd.atlas.2024-08-05+json': {
-                    $ref: '#/components/schemas/Schema',
+                    schema: {
+                      $ref: '#/components/schemas/Schema',
+                    },
                   },
                 },
               },
@@ -214,7 +221,7 @@ testRule('xgen-IPA-117-request-response-bodies-must-be-well-defined', [
     errors: [
       {
         code: 'xgen-IPA-117-request-response-bodies-must-be-well-defined',
-        message: 'Request and response bodies must be well-defined, i.e. include a schema or example(s).',
+        message: 'Request and response bodies must have a schema.',
         path: [
           'paths',
           '/resource/{id}:customMethod',
@@ -227,7 +234,7 @@ testRule('xgen-IPA-117-request-response-bodies-must-be-well-defined', [
       },
       {
         code: 'xgen-IPA-117-request-response-bodies-must-be-well-defined',
-        message: 'Request and response bodies must be well-defined, i.e. include a schema or example(s).',
+        message: 'Request and response bodies must have a schema.',
         path: [
           'paths',
           '/resource/{id}:customMethod',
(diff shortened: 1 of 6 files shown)


tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
```diff
@@ -0,0 +1,266 @@
...
+            },
+          },
+        },
+        '/resource/{id}': {
+          post: {
+            responses: {
+              200: {
+                content: {
+                  'application/vnd.atlas.2024-08-05+json': {
+                    schema: {},
```
yelizhenden-mdb: I think we should also check that the `schema` field refers to a schema instead of being empty

yelizhenden-mdb: Having said that, I could not understand how the `post` method is applicable for `/resource/{id}`

tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
```diff
@@ -0,0 +1,266 @@
+import testRule from './__helpers__/testRule';
+import { DiagnosticSeverity } from '@stoplight/types';
+
+testRule('xgen-IPA-117-request-response-bodies-must-be-well-defined', [
+  {
+    name: 'valid requests and responses',
+    document: {
+      paths: {
+        '/resource/{id}:customMethod': {
```
yelizhenden-mdb, line 9: Could you add a test case for custom method valid Get response as well?

tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
lovisaberggren: > I think we should also check that the schema field refers to a schema instead of being empty

Sure, I can add a check to not only verify schema, example, properties etc. is present, but that it's non-empty

> Having said that, I could not understand how the post method is applicable for /resource/{id}

Ah yeah it's not relevant for this rule, although in general not valid for post requests. I can update just for clarity

tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
lovisaberggren: Sure!

yelizhenden-mdb: Thanks!

tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
```diff
@@ -0,0 +1,301 @@
...
+              401: {
+                $ref: '#/components/responses/Error',
+              },
+              500: {
+                $ref: '#/components/responses/Error',
+              },
+            },
+            requestBody: {
+              content: {
+                // Valid schema
```
yelizhenden-mdb, line 79: Do you have a reference I could take a look at? I checked the [Swagger docs](https://swagger.io/docs/specification/v3_0/adding-examples/) but couldn’t find clarity on if an `example` alone is considered valid

tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
yelizhenden-mdb: Thanks!

wtrocki (approved):

tools/spectral/ipa/__tests__/IPA117RequestResponseBodiesMustBeWellDefined.test.js
lovisaberggren: Hmm good question. I was thinking it is, but it would be pretty rare: it might be more useful for other response types like csv etc. Looking at our current API though, we usually would have something like:
```
"application/vnd.atlas.2023-01-01+csv": {
  "example": "some string response example",
  "schema": {
    "type": "string"
  }
}
```
So we still have schema as type string.

Seems it's not required in the Swagger annotation, and I was looking at the media type object in the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/3.0.4/versions/3.0.4.md#media-type-object), and it doesn't specifically call out any required fields. At the same time, there are no examples provided where there is no schema property defined. 

I _think_ we can probably tighten the rule here to require the schema to be present, I think we'll get the same result for our spec, but I'll experiment a bit.

yelizhenden-mdb: That makes sense, thanks for checking. I think that aligns with my intuition as well. While an example-only media type might technically be allowed, it feels pretty rare in practice. 

That’s why having an example without any schema feels a bit odd to me semantically, even if the spec doesn’t explicitly forbid it. I agree that tightening the rule to require a schema sounds reasonable

lovisaberggren: Updated now, let me know what you think!

matt-condon (approved):

PR #1233, opened 2026-04-17
feat: add operations resource to mcp server

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-393142

The operations resource (openapi://specs/{alias}/operations/{operationId}) returns full detail for a single API operation.

Version filtering is the core feature. The ?version query parameter controls which content is returned:
- latest (default) — only the most recent stable date version
- YYYY-MM-DD — exactly that date version (error if not found)
- preview / upcoming — only those stability levels
- all — every available content type

What gets filtered vs. passed through:
- The 200 response and request body carry versioned Atlas content types (application/vnd.atlas.YYYY-MM-DD+json), so version filtering applies to them
- Non-200 responses (202, 4xx, 5xx) don't participate in versioning — their content is always included as-is

What the response includes:
- Operation metadata: operationId, method, path, summary, description, parameters
- Version metadata: availableVersions, latestStableVersion, hasPreview, hasUpcoming
- Filtered requestBody with schema refs
- All responses with status code, description, and content (200 version-filtered, others raw)

## Checklist

### Changes to Spectral


tools/mcp-server/internal/resources/resources.go
```diff
@@ -35,6 +35,33 @@ func Register(server *mcp.Server, reg *registry.Registry) {
 			"Use this to understand the scope of a spec before searching or slicing it.",
 		MIMEType: mimeTypeJSON,
 	}, makeAliasHandler(reg))
+
+	server.AddResourceTemplate(&mcp.ResourceTemplate{
+		URITemplate: "openapi://specs/{alias}/tags/{tagName}",
```
yelizhenden-mdb, line 40: FYI, this was missed in the previous PR that introduced the resource

tools/mcp-server/internal/resources/uri.go
```diff
@@ -30,6 +30,28 @@ func aliasFromURI(uri string) (string, error) {
...
+
+	// path: /{alias}/operations/{operationId}
+	parts := strings.SplitN(strings.TrimPrefix(u.Path, "/"), "/", 3)
+	if len(parts) != 3 || parts[0] == "" || parts[1] != "operations" || parts[2] == "" {
+		return "", "", "", fmt.Errorf("invalid resource URI %q: expected openapi://specs/{alias}/operations/{operationId}", uri)
+	}
+
+	operationID, err = url.PathUnescape(parts[2])
+	if err != nil {
+		return "", "", "", fmt.Errorf("invalid operationId in URI %q: %w", uri, err)
```
maks-m-mongo-leaf, line 50: Nice usage of %w

tools/mcp-server/internal/resources/operations_helpers.go
```diff
@@ -0,0 +1,198 @@
...
+		ref := op.Responses.Value(code)
+		if ref == nil || ref.Value == nil {
+			continue
+		}
+		desc := ""
+		if ref.Value.Description != nil {
+			desc = *ref.Value.Description
+		}
+		var content []ContentEntry
+		if code == "200" {
```
maks-m-mongo-leaf, line 144: Why do we only apply filters on 200?

tools/mcp-server/internal/resources/operations.go
```diff
@@ -0,0 +1,108 @@
...
+
+func handleOperation(reg *registry.Registry, req *mcp.ReadResourceRequest) (*mcp.ReadResourceResult, error) {
+	alias, operationID, versionFilter, err := aliasAndOperationFromURI(req.Params.URI)
+	if err != nil {
+		return nil, err
+	}
+
+	entry, err := reg.GetByAlias(alias)
+	if err != nil {
+		return nil, fmt.Errorf("spec with alias %q not found", alias)
```
maks-m-mongo-leaf: Would it be reasonable to bubble the error up here?

tools/mcp-server/internal/resources/operations.go
```diff
@@ -0,0 +1,108 @@
...
+	Name        string `json:"name"`
+	In          string `json:"in"`
+	Description string `json:"description,omitempty"`
+	Required    bool   `json:"required"`
+}
+
+// ContentEntry represents a single versioned content type present in a request body or response.
+type ContentEntry struct {
+	ContentType string              `json:"contentType"`
+	Version     string              `json:"version"`
```
maks-m-mongo-leaf: Should we omit empty since this field is not set in `rawContentEntries`?

tools/mcp-server/internal/resources/operations_helpers.go
yelizhenden-mdb: Actually I assumed that the versioning only applies to `200 OK` response and the request body. However, I see that the responses with other status codes have versioning information. 

Nice catch! I will be looking around to understand the situation

tools/mcp-server/internal/resources/operations.go
yelizhenden-mdb: Good point. Instead of duplicating the error message, I think it’s more efficient to delegate it directly. I’ll update the code to bubble up the original error message instead

PR #1235, opened 2026-04-20
fix: resolve stack overflow with circular schema references

andmatei:
## Summary

Fixes a stack-overflow crash in `foascli slice` on OpenAPI specs that contain circular `$ref` schema references (discriminator/inheritance patterns like `CloudProviderContainer <-> AWSCloudProviderContainer`). Upstream root cause: [getkin/kin-openapi#1083](https://github.com/getkin/kin-openapi/issues/1083) — `Validate()` → `visitJSON` walks schema bodies without cycle detection.

The actual fix is small. Drop the warning-only `Validate()` calls from `foascli slice` and `foascli split` — they only emitted `log.Printf("[WARN] ...")` and never affected output. `slice`'s call was the one crashing.

While here, also generalize the JSON-roundtrip duplicate helper that was private to the filter package and use it in `slice` so the loaded spec object is no longer mutated in place by `slice.Slice`.

## What changed

- **`fix(slice)`**: removed `Validate()` from `foascli slice`'s `Run()` — fixes the reported crash.
- **`chore(split)`**: removed the equivalent warning-only `Validate()` from `foascli split` for consistency.
- **`feat(openapi)`**: added `internal/openapi/oasutil.Duplicate` — exported helper for an independent JSON-roundtrip copy. Documented contract: `$ref.Value` is nil on the result by design (kin-openapi serialization doesn't preserve resolved refs across marshal/unmarshal).
- **`refactor(filter)`**: replaced the private `filter.duplicateOas` (and its dedicated test) with the new shared `oasutil.Duplicate`.
- **`refactor(slice)`**: `foascli slice` now duplicates the loaded spec via `oasutil.Duplicate` before passing it to `slice.Slice` (which mutates in place), so the spec object loaded by the caller is left unmodified.
- **`test(slice)`**: added a regression guard that `Run()` leaves its input file byte-identical.

The previous attempt on this branch (`f6ef690618`) had been replaced; the new history starts with a `revert:` of that commit.

> **Note on `oasutil` location:** the shared helper lives at `internal/openapi/oasutil` rather than `internal/openapi` directly — `internal/openapi/versions.go` already imports `internal/openapi/filter`, so `internal/openapi` cannot itself be a dependency of `filter` without an import cycle. A small leaf package side-steps that.

## Test plan

- [x] `go test -race ./internal/...` passes (unit + race detector).
- [x] New unit tests for `oasutil.Duplicate` cover: inline schemas survive with `Value` populated, mutation isolation, refs are nil after round-trip (the documented contract), nil input returns an error.
- [x] `TestOptsRunDoesNotMutateInputFile` regression guard for `foascli slice`.
- [x] End-to-end: `./bin/foascli slice -s openapi/v2.json -o /tmp/clusters-slice.json --tags Clusters` runs to completion (was the crash repro). Output has 16 paths, all tagged `Clusters` only. Input file SHA byte-identical before and after.

### DIFF
--- a/tools/cli/internal/cli/slice/slice.go
+++ b/tools/cli/internal/cli/slice/slice.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/mongodb/openapi/tools/cli/internal/cli/flag"
 	"github.com/mongodb/openapi/tools/cli/internal/cli/usage"
 	"github.com/mongodb/openapi/tools/cli/internal/openapi"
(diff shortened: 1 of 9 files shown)


wtrocki:
is this still relevant?

PR #1243, opened 2026-04-22
feat: add to and from flags to FOASCLI sunset diff command

lovisaberggren:
## Proposed changes

Adds `to` and `from` flags to the `foascli sunset diff` command to filter out endpoints with sunset dates that are outside the range. The flags can also be provided alone/independently. 

The output contains the endpoint and it's sunset date for the base spec and the comparison spec. The endpoint is included in the output if _one of_ the sunset dates are within the provided range.

Adding this feature to be able to create a GH action diff report for sunset drifts where one of the dates are set to "within 3 weeks from now".

_Jira ticket:_ [CLOUDP-397437](https://jira.mongodb.org/browse/CLOUDP-397437)

### DIFF
--- a/tools/cli/internal/cli/sunset/diff.go
+++ b/tools/cli/internal/cli/sunset/diff.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/mongodb/openapi/tools/cli/internal/cli/flag"
 	"github.com/mongodb/openapi/tools/cli/internal/cli/usage"
@@ -35,6 +36,10 @@ type DiffOpts struct {
 	specPath   string
 	outputPath string
 	format     string
+	from       string
+	to         string
+	toDate     *time.Time
+	fromDate   *time.Time
 }
 
 type Diff struct {
@@ -68,7 +73,10 @@ func (o *DiffOpts) Run() error {
 	specSunsets := sunset.NewListFromSpec(specInfo)
 
 	// Find differences
-	var diffs = findDiffs(baseSunsets, specSunsets, o.basePath, o.specPath)
+	diffs, err := o.findDiffs(baseSunsets, specSunsets)
+	if err != nil {
+		return err
+	}
 
 	// Write to output
 	bytes, err := o.newSunsetDiffBytes(diffs)
@@ -84,7 +92,7 @@ func (o *DiffOpts) Run() error {
 	return nil
 }
 
-func findDiffs(baseSunsets, specSunsets []*sunset.Sunset, baseSpecPath, specPath string) []*Diff {
+func (o *DiffOpts) findDiffs(baseSunsets, specSunsets []*sunset.Sunset) ([]*Diff, error) {
 	// Create maps for easy lookup
 	baseMap := make(map[string]*sunset.Sunset)
 	for _, s := range baseSunsets {
@@ -106,15 +114,14 @@ func findDiffs(baseSunsets, specSunsets []*sunset.Sunset, baseSpecPath, specPath
 		if specSunset, exists := specMap[key]; exists {
 			// Endpoint exists in both specs
 			if baseSunset.SunsetDate != specSunset.SunsetDate {
-				// Different sunset dates
 				diffs = append(diffs, &Diff{
 					Operation:      baseSunset.Operation,
 					Path:           baseSunset.Path,
 					Version:        baseSunset.Version,
 					BaseSunsetDate: baseSunset.SunsetDate,
 					SpecSunsetDate: specSunset.SunsetDate,
-					BaseSpec:       baseSpecPath,
-					Spec:           specPath,
+					BaseSpec:       o.basePath,
+					Spec:           o.specPath,
 					Team:           baseSunset.Team,
 				})
 			}
@@ -126,8 +133,8 @@ func findDiffs(baseSunsets, specSunsets []*sunset.Sunset, baseSpecPath, specPath
 				Version:        baseSunset.Version,
 				BaseSunsetDate: baseSunset.SunsetDate,
 				SpecSunsetDate: "",
-				BaseSpec:       baseSpecPath,
-				Spec:           specPath,
+				BaseSpec:       o.basePath,
+				Spec:           o.specPath,
 				Team:           baseSunset.Team,
 			})
 		}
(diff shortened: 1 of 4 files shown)


tools/cli/test/e2e/cli/sunset_test.go
```diff
@@ -13,9 +13,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestDiff_NoChanges(t *testing.T) {
+func TestSunsetDiff_NoChanges(t *testing.T) {
 	baseSpecPath := "../../data/base_spec.json"
-	outputPath := "diff.json"
+	outputPath := getOutputFolder(t, "sunset") + "/diff.json"
```
lovisaberggren: Drive-by fix to write test output into git ignored directory

maks-m-mongo-leaf: [nit] this would make it ever so slightly cleaner https://pkg.go.dev/path#Join

lovisaberggren: Thanks! Done 👍

tools/cli/internal/cli/sunset/diff.go
```diff
@@ -156,7 +163,39 @@ func findDiffs(baseSunsets, specSunsets []*sunset.Sunset, baseSpecPath, specPath
...
+		if err != nil {
+			specSunsetDate = time.Time{} // zero value for time if parsing fails
+		}
+
+		if isDateInRange(&baseSunsetDate, o.fromDate, o.toDate) || isDateInRange(&specSunsetDate, o.fromDate, o.toDate) {
+			out = append(out, d)
+		}
+	}
+
+	return out, nil
```
maks-m-mongo-leaf, line 198: Do we support open intervals here (either of the date is not provided)? If this code returns a nil slice, would something break?
Do we have tests to cover ☝️ ☝️ ?

lovisaberggren: Good Qs!

> Do we support open intervals here (either of the date is not provided)?

Hmm, we do however I don't think it's currently working as expected. When only one of the flags is provided, for example only `from`, I'd expect the behaviour to be "any date from the `from date` forward" and the other way around. I'll add tests to cover this case to ensure it works as expected

> If this code returns a nil slice, would something break?

It wouldn't, it would either be an empty slice or one populated with values. `err` can be an error or nil

lovisaberggren: Made a small adjustment to make it work as I described above, and added tests to cover those cases. Let me know what you think! Ty

tools/cli/internal/cli/sunset/diff.go
```diff
@@ -156,7 +163,39 @@ func findDiffs(baseSunsets, specSunsets []*sunset.Sunset, baseSpecPath, specPath
 		return iKey < jKey
 	})
 
-	return diffs
+	// Filter diffs by date range if specified
+	filteredDiffs, err := o.diffsInRange(diffs)
+	if err != nil {
+		return nil, err
+	}
+
```
saisundar, line 171: `diffsInRange` returns `error` but I don't think any branch ever sets it — the `time.Parse` failures just fall back to `time.Time{}`. Woud it be cleaner to drop the return type and the matching `if err != nil` plumbing in `findDiffs` / `Run()`, or actually propagate a wrapped error when the date string is malformed?

tools/cli/internal/cli/sunset/diff.go
```diff
@@ -156,7 +163,39 @@ func findDiffs(baseSunsets, specSunsets []*sunset.Sunset, baseSpecPath, specPath
...
+	var out []*Diff
+
+	if o.from == "" && o.to == "" {
+		return diffs, nil
+	}
+
+	for _, d := range diffs {
+		baseSunsetDate, err := time.Parse("2006-01-02", d.BaseSunsetDate)
+		if err != nil {
+			baseSunsetDate = time.Time{} // zero value for time if parsing fails
```
saisundar: Nit thing I wanted to check — if someone ships a spec with a malformed sunset date, this coerces to `time.Time{}` and silently drops it from the drift report. Since the whole point of the followup GH action is to alert on drift, a bad date is exactly what we'd want to catch. Maybe gaurd empty-string explicitly and return an error on genuinely malformed non-empty values?

tools/cli/internal/cli/sunset/diff.go
```diff
@@ -186,6 +225,26 @@ func (o *DiffOpts) newSunsetDiffBytes(diffs []*Diff) ([]byte, error) {
 	return yamlData, nil
 }
 
+func (o *DiffOpts) validate() error {
```
saisundar, line 239: Quick one: `validate()` doesn't check that `from <= to`. An inverted range (someone swaps the flags in the workflow YAML) will just return an empty diff, which looks the same as "no drift" and the CI job passes silently. Worth an early error like `--from must not be after --to`?

tools/cli/internal/cli/sunset/list.go
```diff
@@ -94,7 +94,7 @@ func (o *ListOpts) newSunsetInRange(sunsets []*sunset.Sunset) ([]*sunset.Sunset,
 }
 
 func isDateInRange(date, from, to *time.Time) bool {
```
saisundar, line 96: Should this zero-time guard live in `diffsInRange` instead? `isDateInRange` is shared with `sunset list`, which doesn't produce `time.Time{}` sentinels — so the two commands' semantics are now coupled through a case only `diff` creates. Also noticed there's no `list_test.go` change exercising the new branch.

tools/cli/internal/cli/sunset/diff.go
lovisaberggren: Good catch, this is probably left from some changes I made. 

I think we would want to distinguish between the date beling Nil/empty, and invalid. In the case of invalid we should probably return an error instead of silently fail. I'll make some changes 👍

tools/cli/internal/cli/sunset/diff.go
lovisaberggren: We do check that the sunset dates are valid in MMS CI, so it would be caught before the change is shipped to OpenAPI. As per my comment above I'll update to error for invalid dates anyways, if it for some reason would slip through

tools/cli/internal/cli/sunset/diff.go
lovisaberggren: Good point, I'll add a check 👍

tools/cli/internal/cli/sunset/list.go
lovisaberggren: Makes sense, I'll adjust that to be handled in `diffsInRange`

maks-m-mongo-leaf (approved):

PR #1247, opened 2026-04-28
ci: fix FOASCLI goreleaser

### LINKED ISSUE #1246: (prod) The FOASCLI Release has failed. :scream_cat:
See https://github.com/mongodb/openapi/actions/runs/25047180515

lovisaberggren:
_Jira ticket:_ [CLOUDP-400608](https://jira.mongodb.org/browse/CLOUDP-400608)

Fixes #1246 

## Proposed changes

The [previous fix](https://github.com/mongodb/openapi/pull/1245) of removing `mod: mod` from the goreleaser config did not resolve the release failure. The same error persisted:

```
• loading go mod information
  ⨯ release failed after 3s                         
    error=
    │ failed to get module path: exit status 1: go: -mod may only be set to readonly or vendor when in workspace mode, but it is set to “mod”
    │ 	Remove the -mod flag to use the default readonly value, 
    │ 	or set GOWORK=off to disable workspace mode.
```

Reverting previous change and falling back to the other suggested option from the error log, disabling workspace mode for the gorelease runner. Workspace mode is mostly a local development convenience for developing modules depending on each other. Since the FOASCLI does not depend on the mcp-server module, releasing with workspace mode disabled should be safe.

Drive-by: Ran `go work sync` which reordered the `go.work` file and updated some deps.

### Testing

[Release run successful](https://github.com/mongodb/openapi/actions/runs/25058798808)

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -71,6 +71,7 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GOWORK: off
 
   failure-handler: 
     name: Failure Handler
--- a/tools/mcp-server/go.mod
+++ b/tools/mcp-server/go.mod
@@ -7,10 +7,12 @@ require (
 	github.com/modelcontextprotocol/go-sdk v1.5.0
 	github.com/mongodb/openapi/tools/cli v0.0.0
 	github.com/spf13/afero v1.15.0
+	github.com/stretchr/testify v1.11.1
 )
 
 require (
 	cloud.google.com/go v0.123.0 // indirect
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/go-openapi/jsonpointer v0.22.5 // indirect
 	github.com/go-openapi/swag/jsonname v0.25.5 // indirect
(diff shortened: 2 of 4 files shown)


.github/workflows/release-cli.yml
```diff
@@ -71,6 +71,7 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
andreaangiolillo, line 73: Can we revert the changes in https://github.com/mongodb/openapi/pull/1245?

lovisaberggren: Sure, done!

andreaangiolillo (approved):
LGTM

tools/mcp-server/go.mod
```diff
@@ -7,10 +7,12 @@ require (
 	github.com/modelcontextprotocol/go-sdk v1.5.0
 	github.com/mongodb/openapi/tools/cli v0.0.0
 	github.com/spf13/afero v1.15.0
+	github.com/stretchr/testify v1.11.1
```
andreaangiolillo, line 10: why are these needed?

lovisaberggren: These were added by `go work sync`, which syncs dependencies for the modules declared in the `go.work` file. Same result as running `go mod tidy` inside the `mcp-server` directory. I think these were missing, but are used

PR #1284, opened 2026-05-22
feat: add cleanup process for sunsetted collections

yelizhenden-mdb:
## Proposed changes

_Jira ticket:_ CLOUDP-260980

**Problem**
     
Every time a new Atlas Admin API version is released, a new Postman collection is published but the old one is only renamed  (⭐ removed) and left in the workspace. Over time, this accumulates redundant collections that create noise for customers browsing the workspace.
  
The Postman collection is generated from the latest FOAS, which contains all endpoints across all API versions. Customers select the API version they want via the Accept header, not by choosing a different collection. Old collections don't provide any additional endpoint coverage.
  
  **What changed**
  
`  tools/postman/scripts/upload-collection.sh
`  
  - Removed the PATCH call that renamed the previous starred collection (removing ⭐  from its name) before creating a new one
  - Added a cleanup loop at the end of the script that deletes all MongoDB Atlas Administration API collections from the workspace except the current one
  
The cleanup runs after every upload. Because the initial collection list is fetched before the new collection is posted, the current collection is safely excluded by name regardless of whether it was just created or updated in place.

# Testing
Created new test collections with old versions on Postman:
Screenshot 2026-05-25 at 11 43 36
After running the script, successfully confirmed the old collections are removed:
Screenshot 2026-05-25 at 11 46 27

## Checklist

### Changes to Spectral


tools/postman/scripts/upload-collection.sh
```diff
@@ -111,4 +91,29 @@ else
...
+    --header "X-API-Key: ${POSTMAN_API_KEY}")
+  if [[ "${http_code}" != "200" ]]; then
+    echo "[ERROR] Failed to delete old collection: ${name} (id: ${id}), HTTP status: ${http_code}"
+  else
+    deleted=$((deleted + 1))
+  fi
+done < <(jq -c --arg current "${current_collection_name}" \
+  '.collections[] | select(.name | contains("MongoDB Atlas Administration API")) | select(.name != $current)' \
+  "${COLLECTIONS_LIST_FILE}")
+echo "[SUMMARY] Deleted ${deleted} old collection(s)"
```
andreaangiolillo, line 117: is this a breaking changes for customers using that collection? Do you remember the reasons why we decided to rename the collection?

yelizhenden-mdb: The customers always must use the latest collection. Can you think of any use case a customer might need to use a previous collection? 
This PR does not introduce any renaming, it is just cleaning up the previous collections

andreaangiolillo: > Can you think of any use case a customer might need to use a previous collection?

I don't have enough context on how we generate the collection. In particular, it would be helpful to understand how we handle deprecated versions. Are we removing them from the collection and requiring customers to move to the new API version? If so, one use case for keeping the old collection available would be to let customers continue using the deprecated endpoints until they've fully migrated.

> This PR does not introduce any renaming, it is just cleaning up the previous collections

Got it. My question is more along the lines of: if a customer is currently using any of the collections you're planning to delete, will they encounter an error? If so, that wouldn't be a great experience, and I'm wondering if that was the reason we chose to rename the old collection when we worked on Postman instead of deleting them.

yelizhenden-mdb: Valid concerns. Here is my understanding of how this works:

- The collection is generated from the latest FOAS (`openapi-YYYY-MM-DD.json` file). Deprecated endpoints will still be included as long as they haven't been sunset
- If a customer forks the collection, it persists locally and independently, so deleting the source will have no impact. If a customer references the collection directly by URL or ID, deletion could cause an issue
- I implemented the star feature as well. The original requirement was just to make the latest collection more prominent. I haven't been asked to remove the older ones, or that option wasn't considered at the time

Also, to ensure I didn't misinterpret the ticket, I'm going to sync with Stephen to confirm this PR aligns with what he wanted. Thanks for calling this out!

yelizhenden-mdb: @stephen-cassil0 confirmed that we only want to list the most recent collection. He also noted that deleting older collections and any potential side effects are not an issue

andreaangiolillo (approved):

PR #1286, opened 2026-05-26
ci: push Go submodule tag on release

yelizhenden-mdb:
## Summary
- Adds a second tag (`tools/cli/vX.Y.Z`) during the cli release workflow so the Go module is resolvable from outside this repo.
- `tools/cli/v0.0.65` has already been backfilled manually so external consumers can pin the current latest version immediately.

Jira: CLOUDP-407910

## Test plan
- [x] `tools/cli/v0.0.65` resolves via the Go proxy (`go list -m github.com/mongodb/openapi/tools/cli@v0.0.65`).
- [ ] Next release run pushes both tags (verified on first post-merge release).

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -37,6 +37,13 @@ jobs:
           commit_sha: ${{ steps.get-sha.outputs.sha }}
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg_passphrase: ${{ secrets.PASSPHRASE }}
+      - name: Create Go submodule tag for tools/cli
+        uses: rickstaa/action-create-tag@a1c7777fcb2fee4f19b0f283ba888afa11678b72
+        with:
+          tag: tools/cli/${{ inputs.version_number }}
+          commit_sha: ${{ steps.get-sha.outputs.sha }}
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg_passphrase: ${{ secrets.PASSPHRASE }}
 
   run-tests:
     needs: [ create-tag ]


.github/workflows/release-cli.yml
```diff
@@ -37,6 +37,13 @@ jobs:
           commit_sha: ${{ steps.get-sha.outputs.sha }}
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           gpg_passphrase: ${{ secrets.PASSPHRASE }}
+      - name: Create Go submodule tag for tools/cli
+        uses: rickstaa/action-create-tag@a1c7777fcb2fee4f19b0f283ba888afa11678b72
+        with:
+          tag: tools/cli/${{ inputs.version_number }}
```
andreaangiolillo, line 43:
```suggestion
          tag: tools/foas/${{ inputs.version_number }}
```

If we want to expose the logic of foascli as a module, I think we should avoid using "CLI" in the name, since it implies a command line interface rather than a module.

WDYT?

andreaangiolillo (approved):
my comment is not blocking. LGTM

.github/workflows/release-cli.yml
yelizhenden-mdb: Fair point. The cli in the path comes from where the code lives today (`tools/cli/ `is the home of `foascli`, and the library lives there too).
 
One option is to rename `tools/cli` to `tools/foas`. The directory would still hold both the binary and the library, just under a clearer name. From what I can see, this only needs small updates in `oasis` and openapi CI/release config, no external consumers break (e.g. MMS uses the binary, not the Go module).
 
Or, we can create a new `tools/foas` module without touching the existing `tools/cli`, and move only the public packages there. WDYT?

andreaangiolillo: I'd like much more the second option if possible so that both tools (cli and MCP) will reuse the same module. This is a medium redesign so feel free to create a ticket for it so that we can triage it

PR #1294, opened 2026-05-27
refactor: split into tools/foas (library) and tools/cli (binary)

yelizhenden-mdb:
## Summary

Splits `tools/cli` into two Go modules and splits the corresponding CI/release workflows:

- **`tools/foas/`** — the library. Public packages: `openapi`, `apiversion`, `breakingchanges`, `changelog`, `pointer`. Consumed by mcp-server (in `mongodb-labs/oasis`) and future Go projects.
- **`tools/cli/`** — the binary (`foascli`). Cobra commands, entry point, version package. Depends on `tools/foas` via a `replace` directive in `go.mod`.

The library's `internal/` boundary is removed so consumers import packages directly. The `pkg/` facade wrappers are deleted (no longer needed).

Workflows are split per module so library and binary release on independent cadences (kubernetes/client-go style).

## New directory layout

```
tools/foas/                       # LIBRARY MODULE
├── go.mod                        # github.com/mongodb/openapi/tools/foas
├── go.sum
├── Makefile                      # library-only: build, test, lint, fmt, gen-docs
├── .golangci.yml
├── README.md, RELEASING.md
├── apiversion/                   # version parsing, stability levels
├── breakingchanges/              # breaking-change detection, exemptions
├── changelog/                    # changelog generation
│   └── outputfilter/
├── openapi/                      # spec loading, slicing, save
│   ├── errors/
│   ├── filter/
│   │   └── template/
│   ├── slice/
│   └── sunset/
├── pointer/                      # pointer utility
├── scripts/                      # doc generation
└── test/data/                    # shared test fixtures (used by both modules)

tools/cli/                        # BINARY MODULE  (depends on ../foas via replace)
├── go.mod                        # github.com/mongodb/openapi/tools/cli
├── go.sum
├── Makefile                      # binary build/install/release pipeline
├── .goreleaser.yml               # ldflags -> tools/cli/version
├── .golangci.yml
├── README.md, RELEASING.md, precommit.sh
├── cmd/foascli.go                # binary entry point
├── version/                      # binary build version (ldflag target)
├── breakingchanges/              # cobra command builder
│   └── exemptions/
├── changelog/                    # cobra command builder
│   ├── convert/
│   └── metadata/
├── filter/, flag/, merge/, slice/, split/, sunset/, usage/, validator/, versions/   # cobra commands
├── root/openapi/                 # root cobra command
└── test/                         # test helpers + e2e
    ├── cmdvalidator.go
    └── e2e/
```

## Workflows

| File | Trigger | Purpose |
|---|---|---|
| `code-health-foas.yml` | push/PR on `tools/foas/**` | Build + test + lint for the library |
| `code-health-cli.yml` | push/PR on `tools/cli/**` or `tools/foas/**` | Build + test + lint + e2e for the binary (rebuilds when library changes since cli uses local foas via `replace`) |
| `release-foas.yml` | manual dispatch | Tag `tools/foas/vX.Y.Z`, run library QA. No binary. |
| `release-cli.yml` | manual dispatch | Tag `vX.Y.Z` + `tools/cli/vX.Y.Z`, run cli QA, GoReleaser publishes the binary. |

The old matrixed `code-health-foascli.yml` is replaced by the two new workflows.

## Unchanged for external consumers

- Binary name: `foascli`
- Release archive name: `mongodb-foas-cli_*.tar.gz`
- mms's Bazel-downloaded archives continue to work unchanged.

## Why

- The `cli` segment in the module path was misleading — mcp-server consumes the library, not the CLI.
- Splitting into two modules lets library and binary version independently and keeps library consumers from inheriting cobra/viper transitively.
- Library packages out of `internal/` make the public API the real public API, not a hand-curated facade.
- Splitting workflows means library-only PRs don't run cli builds and library releases don't pull cli into the gating tests.

Jira: CLOUDP-407910

## Test plan

- [x] `go build ./...` + `go test ./...` pass in `tools/foas/`.
- [x] `go build ./...` + `make unit-test` pass in `tools/cli/` (e2e runs separately via `make e2e-test`).
- [x] No remaining `tools/cli/internal`, `tools/cli/pkg`, or stale path references.

## Follow-ups (separate PRs)

- First release after merge: cut a foas library release (`tools/foas/vX.Y.Z`) so oasis can pin it.
- oasis (`mongodb-labs/oasis`) updates its `go.mod` to require `tools/foas` at the new version and rewrites imports.
- Optional: cli release whenever there's a meaningful binary change for users (mms, etc.).


tools/foas/scripts/doc_filters.sh
```diff
@@ -19,4 +19,4 @@ echo " - Filtering per version, so that only the endpoints that are available in
 echo "## What filters are available?"
 
 echo "### List of filters"
-grep -n '// .*Filter: .*' internal/openapi/filter/*.go | sort -u -k2 | sed -n "s/\([^0-9]*\):\([0-9]*\):\/\/ \(.*Filter: .*\)/[\3](\.\.\/\1?plain=1#L\2)  /p" | sort
```
maks-m-mongo-leaf, line 22: I am struggling to assess the effect of this change.

yelizhenden-mdb: It is just renaming, there is no `internal` directory anymore

yelizhenden-mdb: About the changes in `sed` pattern: it is again some changes supporting to the new layout and some refinements on capturing the filename and how to locate the files

andreaangiolillo (commented):
I left one comment

.github/workflows/code-health-cli.yml
```diff
@@ -1,17 +1,22 @@
-name: 'Code Health FoasCLI'
+name: 'Code Health CLI'
```
andreaangiolillo, line 1: This PR is quite large, which makes it a bit tricky to review thoroughly. Would it be possible to split it into smaller pieces of work and merge them into a feature branch first? That would help avoid any unintended impact on the current FOAS release process. One possible way to break it down:

1. Add the new structure for tool/foas as a clean Go project.
2. Move a single package from the CLI project into the FOAS module, import it, and see how it behaves.
3. Move the remaining shared logic into the FOAS module.
4. Update the GitHub Actions files.
5. Merge the feature branch into main.


This is just a suggestion, but going step by step should give you more confidence that the changes work as expected and that we are not introducing any regressions along the way. Also we don't need a feature branch if you don't think is useful in this case. 

One general piece of feedback: I think we should keep the `cli/internal` directory and place all the CLI commands inside it, since they are not meant to be exported.

yelizhenden-mdb: Sounds good to me, let me convert this PR to draft and go with the plan as you suggested.

> One general piece of feedback: I think we should keep the cli/internal directory and place all the CLI commands inside it, since they are not meant to be exported.

Just to confirm my understanding: you are aligned with the new splitted layout but you think that the code should live under an extra `internal` directory under `tools/cli`. Am I correct?

PR #1309, opened 2026-06-15
chore: move apiversion package into tools/foas

yelizhenden-mdb:
## What this introduces

**Branch 2 of 4** in splitting `tools/cli` into a library (`tools/foas`) + binary (`tools/cli`). Builds on #1308 (now merged). This is the **pilot move**: it relocates one package — `apiversion` — to prove out the cross-module wiring before moving the rest.

### Changes

- **Moved** `tools/cli/internal/apiversion/` → `tools/foas/apiversion/` (4 files: `version.go`, `stabilitylevel.go`, + tests). Package name unchanged (`package apiversion`).
- **Rewrote imports** in all **20** cli callers from `…/tools/cli/internal/apiversion` to `…/tools/foas/apiversion`.
- **Deleted** `tools/cli/pkg/apiversion/` — a thin re-export wrapper with **zero importers** (dead code; the whole point of `tools/foas` is to be the public surface).
- **Wired the module dependency** in `tools/cli/go.mod`:
  ```
  require github.com/mongodb/openapi/tools/foas v0.0.0
  replace github.com/mongodb/openapi/tools/foas => ../foas
  ```
  `go.work` resolves foas locally during dev/CI; the `replace` covers `GOWORK=off` (GoReleaser).
- **`go mod tidy`** surfaced that `apiversion` (and its tests) pull in `kin-openapi` and `testify`, so `tools/foas` now declares those directly — hence the new `tools/foas/go.sum` and regenerated `tools/go.work.sum`.

### Verification

This PR only **moves** the `apiversion` package (no logic change; package name and behavior unchanged; the expected-output specs committed in the repo are untouched). The `foascli` binary is consumed by several release/generation workflows (`release-cli`, `release-spec`, `release-changelog`, `generate-openapi`, `changelog-report`, `required-spec-validations`, `api-versions-reminder`), so behavioral parity is verified end-to-end:

**CI (Linux) — all green on this PR:**
| Check | Result |
|---|---|
| `e2e-tests` (builds `foascli`, runs commands, compares output against the **expected specs committed in the repo**) | ✅ pass |
| build (foascli + foas-lib) | ✅ pass |
| lint (both modules) | ✅ pass |
| unit-tests (ubuntu + macos, both modules) | ✅ pass |

**Local (macOS) — second, independent signal:**
- `apiversion`-dependent e2e tests (`split`, `changelog`, `versions`) → `ok …/test/e2e/cli 157.001s` (exit 0)

[...]

PR #1310, opened 2026-06-16
chore: move breakingchanges, changelog, openapi, pointer to foas module

yelizhenden-mdb:
## Summary

Step 3 of the `tools/cli` → `tools/foas` module split. Extracts the remaining reusable packages from `tools/cli/internal/` into the standalone `tools/foas` library module.

### Previous steps
- Step 1 — scaffold `tools/foas` library module: #1308 (merged)
- Step 2 — pilot move of `apiversion`: #1309 (merged)
- **Step 3 — this PR**: move the remaining packages

## Changes

- **Move** `breakingchanges`, `changelog` (+ `outputfilter`), `openapi` (+ `slice`, `sunset`, `filter`, `errors`), and `pointer` from `tools/cli/internal/` to `tools/foas/`.
- **Delete `tools/cli/pkg/` entirely** — the thin wrappers had no importers and are superseded by the public `foas` module.
- **Rewrite all importers** (42 files) and `go:generate` mockgen directives from `tools/cli/internal/{pkg}` → `tools/foas/{pkg}`.
- **Test fixtures:** the moved packages dropped the `internal/` path segment, so their relative `test/data` paths needed one fewer `../`. Copied the minimal subset foas needs into `foas/test/data/` (~1.6MB) rather than relocating the 225MB changelog corpus that cli's remaining tests still depend on.
- `go mod tidy` on `foas/go.mod` to pull in the new dependencies (oasdiff, afero, mock, strcase, …).

Dependency direction stays one-way: `cli → foas`. `foas` references nothing in `cli`.

## Verification
`Build`, `unit-tests` and `e2e-tests` of CLI passes

## Next

Branch 4 — CI & release wiring (rename `code-health-foascli.yml`, add `release-foas-lib.yml`).


tools/foas/go.mod
```diff
@@ -6,18 +6,32 @@ toolchain go1.26.0
...
+	github.com/oasdiff/oasdiff v1.19.1
+	github.com/spf13/afero v1.15.0
 	github.com/stretchr/testify v1.11.1
+	go.uber.org/mock v0.6.0
+	golang.org/x/text v0.38.0
+	gopkg.in/yaml.v3 v3.0.1
 )
 
 require (
-	github.com/davecgh/go-spew v1.1.1 // indirect
```
andreaangiolillo, line 13: [q] Have we tried releasing the foas module and importing it into the CLI? We want to make sure the CLI is always ready to release, in case we have to fix an OpenAPI issue. If we merge this PR without the foas module in the CLI, we'd be blocked from making an emergency release until the last PR of this migration is completed.

yelizhenden-mdb: No, not really. The CLI intentionally consumes the FOAS library via `go.work` / `replace` rather than a published tag, as they are co-developed in the same repo. Since the E2E tests are passing, I don't think there is any change to the CLI build output.

Additionally, there isn't a release workflow for the FOAS library yet, which will be handled as part of Step 4. After that, I plan to use the versioned library in the OpenAPI MCP implementation. Let me know if you have any suggestions or concerns regarding this approach

yelizhenden-mdb: > If we merge this PR without the foas module in the CLI, we'd be blocked from making an emergency release until the last PR of this migration is completed.

Now I see the edited comment. Do you suggest including a published and tagged foas library in the CLI instead of using Go workspaces?

andreaangiolillo: > Do you suggest including a published and tagged foas library in the CLI instead of using Go workspaces?

Yes, but you raised a fair point: the e2e tests are passing, so I assume that, when we build the CLI binary via `make build`, the compiler fetches the foas module even if it isn't released (using go workspace). Feel free to ignore my comment. Thanks!

andreaangiolillo (approved):
LGTM

PR #1311, opened 2026-06-16
chore: wire up CI and release workflows for foas module

yelizhenden-mdb:
## Summary

Step 4 (final) of the `tools/cli` → `tools/foas` module split. Wires up CI and release for the new library module.

### Previous steps
- Step 1 — scaffold `tools/foas` library module: #1308 (merged)
- Step 2 — pilot move of `apiversion`: #1309 (merged)
- Step 3 — move remaining packages: #1310
- **Step 4 — this PR**: CI & release wiring

## Changes

- **Add a `tools/foas/**` path trigger to `code-health-foascli.yml`.** The CLI depends on foas, so its build/test/lint/e2e pipeline should also re-run when the library changes.
- **Add `release-foas-lib.yml`** — a manual-dispatch workflow that releases the foas library. Since foas is a Go library (no binaries) and is consumed as source by the CLI via Go workspaces, a "release" is simply the Go submodule tag `tools/foas/vX.Y.Z`. External repos (e.g. `ipa-skills`) then consume it with:

  ```
  go get github.com/mongodb/openapi/tools/foas@vX.Y.Z
  ```

## Follow-up after merge

Once merged, I'll cut the first library tag by running the release workflow. The resulting tag `tools/foas/v0.1.0` will appear under the repo's Tags - not Releases, since the library has no binary artifact.

### DIFF
--- a/.github/workflows/release-foas-lib.yml
+++ b/.github/workflows/release-foas-lib.yml
@@ -1,11 +1,7 @@
-name: 'Foas Library New Release'
+name: 'FOAS Library New Release'
 run-name: 'Release foas ${{ inputs.version_number }} (skip tests: ${{ inputs.skip_tests }})'
 
-# Used for releasing the tools/foas Go library module. The module is consumed as
-# source by tools/cli (via go.work/replace), so a "release" is simply a Go
-# submodule tag (tools/foas/vX.Y.Z) that external repos can `go get`.
-# Tests run before the tag is created: the tag is immutable on the Go module
-# proxy once published, so we validate the candidate commit first.
+# Used for releasing the tools/foas Go library module.
 on:
   workflow_dispatch:
     inputs:
@@ -17,19 +13,20 @@ on:
         default: 'false'
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   run-tests:
     if: >-
       !cancelled()
       && inputs.skip_tests == 'false'
-    secrets: inherit
     uses: ./.github/workflows/code-health-foas-lib.yml
 
   create-tag:
     runs-on: ubuntu-latest
     needs: [ run-tests ]
+    permissions:
+      contents: write # required to push the release tag
     # Tag is created only when tests pass (or are explicitly skipped).
     if: >-
       !cancelled()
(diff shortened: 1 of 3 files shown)


.github/workflows/release-foas-lib.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Create Go submodule tag for tools/foas
+        uses: rickstaa/action-create-tag@a1c7777fcb2fee4f19b0f283ba888afa11678b72
+        with:
+          tag: tools/foas/${{ inputs.version_number }}
+          commit_sha: ${{ steps.get-sha.outputs.sha }}
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg_passphrase: ${{ secrets.PASSPHRASE }}
+
+  failure-handler:
+    name: Failure Handler
```
andmatei, line 52: Should we have `run-tests` here too? If I understand correctly, `create-tag` fails if the test fail

.github/workflows/release-foas-lib.yml
```diff
@@ -0,0 +1,60 @@
...
+
+permissions:
+  contents: write
+
+jobs:
+  run-tests:
+    if: >-
+      !cancelled()
+      && inputs.skip_tests == 'false'
+    secrets: inherit
```
andmatei: Just double checking if the `code-health-foas-lib` needs any secrets from the GH action?

.github/workflows/release-foas-lib.yml
```diff
@@ -0,0 +1,60 @@
...
+    inputs:
+      version_number:
+        description: 'Version number (e.g., v1.0.0, v1.0.0-pre, v1.0.0-pre1)'
+        required: true
+      skip_tests:
+        description: 'Set value to `true` to skip tests, default is `false`'
+        default: 'false'
+
+permissions:
+  contents: write
```
andmatei: Q: do we need `write` here or is `read` enough?

.github/workflows/release-foas-lib.yml
yelizhenden-mdb: The way I see it, this is a workflow for creating a tag. If `run-tests` fails, `create-tag` will be skipped, which isn't exactly a failure in itself. However, we do need a way to track whether `create-tag` was skipped specifically because of a failure, so adding that visibility would be a great idea

.github/workflows/release-foas-lib.yml
yelizhenden-mdb: Hmm, no not really. Good catch, let me drop that

.github/workflows/release-foas-lib.yml
yelizhenden-mdb: write needed because create-tag step pushes a git tag, and pushing refs needs write. However, I can change it to having write permission only for create-tag job. Good call!

andmatei (approved):
LGTM

PR #1322, opened 2026-06-17
feat(ipa): add MCP to acronym ignore lists

andmatei:
## Summary
- Adds `MCP` to the `ignoreList` in IPA-117 (operation summary format) and IPA-126 (tag names title case) so the acronym is not flagged by casing rules.

## Test plan
- [x] Verify spectral rules pass on specs using `MCP` in operation summaries and tag names


yelizhenden-mdb (approved):

wtrocki (approved):

wtrocki:
Future though: Add ability to support list downstream so we do not need to make upstream fixes with terms.

PR #1406, opened 2026-07-29
fix(ipa): allow Operations endpoint paths in IPA-102 alternation check

julius-jogela:
## Proposed changes

**Problem:** LRO standardization (IPA-132) needs read-only Operations endpoints nested under a parent resource, at both collection and instance scope:
- `<parent>/operations`
- `<parent>/operations/{operationId}`
- `<parent>/{resourceId}/operations`
- `<parent>/{resourceId}/operations/{operationId}`

`xgen-IPA-102-path-alternate-resource-name-path-param` rejects the two collection-scoped shapes today, because the rule expects path segments to alternate resource-name / path-param, and `operations` sits where a `{param}` is expected.

**Fix:** treat a trailing `operations` or `operations/{operationId}` as a parity-neutral suffix, strip it off the path first, then run the existing alternation check unchanged on what's left.

**Verification:** replaying the old and new rule logic over all 367 paths in `openapi/.raw/v2.yaml` produces zero behavior changes on existing paths, confirming the change is additive.

_Jira ticket:_ [CLOUDP-429009](https://jira.mongodb.org/browse/CLOUDP-429009)

## Checklist
- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [x] I have added tests that prove my fix is effective or that my feature works

### Changes to Spectral
- [x] I have read the [README](../tools/spectral/README.md) file for Spectral Updates

PR #1414, opened 2026-08-04
fix(ipa): preserve resource scope in Operations resource operation IDs

julius-jogela:
## Proposed changes

**Problem:** LRO standardization (IPA-132) nests read-only Operations endpoints under a parent resource at both collection and instance scope. The collection- and instance-scoped pairs generated **identical** operation IDs:

| Path | Operation ID (before) |
| --- | --- |
| `/clusters/operations` | `listGroupClusterOperations` |
| `/clusters/{clusterName}/operations` | `listGroupClusterOperations` |
| `/clusters/operations/{operationId}` | `getGroupClusterOperation` |
| `/clusters/{clusterName}/operations/{operationId}` | `getGroupClusterOperation` |

`generateOperationID` strips all path parameters before building the noun list and then singularizes every non-final noun, so both scopes collapse to the same nouns. Note this affects **every** method on these paths, not just List — the `{operationId}` parameter does not make Get naturally distinct, because it is discarded along with the other path parameters.

**Fix:** add an `isCollectionScopedOperationsPath` check and keep the parent noun in its plural form when the trailing `operations` section is attached to a resource collection rather than to a single resource. The trailing noun keeps its normal meaning (plural for List, singular for Get), so the existing convention is preserved and all four IDs become distinct:

| Path | Operation ID (after) |
| --- | --- |
| `/clusters/operations` | `listGroupClustersOperations` |
| `/clusters/{clusterName}/operations` | `listGroupClusterOperations` |
| `/clusters/operations/{operationId}` | `getGroupClustersOperation` |
| `/clusters/{clusterName}/operations/{operationId}` | `getGroupClusterOperation` |

This reuses the existing `singularize` helper and its `ignoreSingularizationList` opt-out rather than introducing new `Collection`/`Instance` suffix conventions.

**Documentation:** the exception is documented in the `xgen-IPA-104-valid-operation-id` and `xgen-IPA-105-valid-operation-id` rule descriptions (the Get and List methods), and the ruleset `README.md` is regenerated to match.

**Verification:** unchanged behavior for non-Operations paths, a singular parent (`/groups/{groupId}/operations`), a non-trailing `operations` section, an unscoped `/api/atlas/v2/operations`, and multi-word custom methods — all covered by tests. Full suite passes (705 tests), along with `lint-js` and `format-check`.

Because the scope distinction is a property of the path rather than of the method, the Create, Update and Delete shapes are pinned by tests as well:

| Path | Operation ID |
| --- | --- |
| `create` `/clusters/operations` | `createGroupClustersOperation` |
| `create` `/clusters/{clusterName}/operations` | `createGroupClusterOperation` |
| `update` `/clusters/operations/{operationId}` | `updateGroupClustersOperation` |
| `update` `/clusters/{clusterName}/operations/{operationId}` | `updateGroupClusterOperation` |
| `delete` `/clusters/operations/{operationId}` | `deleteGroupClustersOperation` |
| `delete` `/clusters/{clusterName}/operations/{operationId}` | `deleteGroupClusterOperation` |

These endpoints **should not exist** — IPA-132 Operations resources are read-only — so the tests are not asserting supported API surface. They are there purely to guarantee the collection-vs-instance distinction stays consistent across every method rather than being special-cased to Get and List, and to catch a silent regression if a mutating method is ever added to an Operations path.

Follows up on #1406, which relaxed `xgen-IPA-102-path-alternate-resource-name-path-param` to permit these paths.

_Jira ticket:_ [CLOUDP-429467](https://jira.mongodb.org/browse/CLOUDP-429467)

## Checklist

### Changes to Spectral

## Further comments

One thing worth flagging for review: there is no operation-ID uniqueness check anywhere in the repo today — each rule only validates that an operation ID matches the one generated from its own path. The collision described above would not have been flagged by validation, only by downstream SDK codegen, where each operation ID becomes a method name.

### DIFF
--- a/tools/spectral/ipa/__tests__/utils/operationIdGeneration.test.js
+++ b/tools/spectral/ipa/__tests__/utils/operationIdGeneration.test.js
@@ -29,6 +29,89 @@ describe('tools/spectral/ipa/utils/operationIdGeneration.js', () => {
       expect(generateOperationID('grant', '/api/atlas/v2/groups/{groupId}/access')).toEqual('grantGroupAccess');
     });
 
+    it('should preserve the resource scope of operations paths', () => {
+      // the collection-scoped paths keep the parent plural, the instance-scoped paths keep it singular,
+      // so that the two Operations resources do not share an operation ID
+      expect(generateOperationID('list', '/api/atlas/v2/groups/{groupId}/clusters/operations')).toEqual(
+        'listGroupClustersOperations'
+      );
+      expect(generateOperationID('list', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations')).toEqual(
+        'listGroupClusterOperations'
+      );
+      expect(generateOperationID('get', '/api/atlas/v2/groups/{groupId}/clusters/operations/{operationId}')).toEqual(
+        'getGroupClustersOperation'
+      );
+      expect(
+        generateOperationID('get', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations/{operationId}')
+      ).toEqual('getGroupClusterOperation');
+    });
+
+    it('should generate distinct operation IDs for collection- and instance-scoped operations resources', () => {
+      expect(generateOperationID('list', '/api/atlas/v2/groups/{groupId}/clusters/operations')).not.toEqual(
+        generateOperationID('list', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations')
+      );
+      expect(
+        generateOperationID('get', '/api/atlas/v2/groups/{groupId}/clusters/operations/{operationId}')
+      ).not.toEqual(
+        generateOperationID('get', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations/{operationId}')
+      );
+    });
+
+    it('should preserve the resource scope for mutating methods on operations paths', () => {
+      // IPA-132 Operations resources are read-only, so these methods should not exist. Pinned to
+      // ensure the scope distinction stays consistent across every method, not just Get and List.
+      expect(generateOperationID('create', '/api/atlas/v2/groups/{groupId}/clusters/operations')).toEqual(
+        'createGroupClustersOperation'
+      );
+      expect(generateOperationID('create', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations')).toEqual(
+        'createGroupClusterOperation'
+      );
+      expect(generateOperationID('update', '/api/atlas/v2/groups/{groupId}/clusters/operations/{operationId}')).toEqual(
+        'updateGroupClustersOperation'
+      );
+      expect(
+        generateOperationID('update', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations/{operationId}')
+      ).toEqual('updateGroupClusterOperation');
+      expect(generateOperationID('delete', '/api/atlas/v2/groups/{groupId}/clusters/operations/{operationId}')).toEqual(
+        'deleteGroupClustersOperation'
+      );
+      expect(
+        generateOperationID('delete', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations/{operationId}')
+      ).toEqual('deleteGroupClusterOperation');
+    });
+
+    it('should not preserve the parent plural for other operations paths', () => {
+      // the parent is a single resource, so the existing singularization applies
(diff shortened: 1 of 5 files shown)


andmatei (approved):
LGTM

yelizhenden-mdb (changes requested):
Waiting for DevTools team confirmation

tools/spectral/ipa/rulesets/IPA-104.yaml
```diff
@@ -104,6 +104,7 @@ rules:
       The Operation ID must start with the verb “get” and should be followed by a noun or compound noun.
       The noun(s) in the Operation ID should be the collection identifiers from the resource identifier in singular form.
       If the resource is a singleton resource, the last noun may be the plural form of the collection identifier.
+      For a collection-scoped Operations resource, such as '/clusters/operations/{operationId}', the parent noun is in plural form, so that the Operation ID differs from the one for the instance-scoped Operations resource, such as '/clusters/{clusterName}/operations/{operationId}'.
```
yelizhenden-mdb: Operation IDs for collection-scoped and instance-scoped LRO resources are already expected to be different — as are all operation IDs, since they map to CLI command names. Restating that in an operation-ID validity check doesn't add value, especially since the rule doesn't actually verify cross-path uniqueness. If you want to document the change, it's enough to say what the rule now expects:

`For a collection-scoped Operations resource, such as /resources/operations/{operationId}, the parent noun (e.g. resources) stays in its plural form instead of being singularized`

yelizhenden-mdb: Same applies to IPA-105

tools/spectral/ipa/__tests__/utils/operationIdGeneration.test.js
```diff
@@ -29,6 +29,89 @@ describe('tools/spectral/ipa/utils/operationIdGeneration.js', () => {
...
+      );
+      expect(generateOperationID('get', '/api/atlas/v2/groups/{groupId}/clusters/operations/{operationId}')).toEqual(
+        'getGroupClustersOperation'
+      );
+      expect(
+        generateOperationID('get', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations/{operationId}')
+      ).toEqual('getGroupClusterOperation');
+    });
+
+    it('should generate distinct operation IDs for collection- and instance-scoped operations resources', () => {
```
yelizhenden-mdb: Do we need this test, given that we already have the previous check?

tools/spectral/ipa/__tests__/utils/operationIdGeneration.test.js
```diff
@@ -29,6 +29,89 @@ describe('tools/spectral/ipa/utils/operationIdGeneration.js', () => {
...
+      expect(generateOperationID('delete', '/api/atlas/v2/groups/{groupId}/clusters/operations/{operationId}')).toEqual(
+        'deleteGroupClustersOperation'
+      );
+      expect(
+        generateOperationID('delete', '/api/atlas/v2/groups/{groupId}/clusters/{clusterName}/operations/{operationId}')
+      ).toEqual('deleteGroupClusterOperation');
+    });
+
+    it('should not preserve the parent plural for other operations paths', () => {
+      // the parent is a single resource, so the existing singularization applies
```
yelizhenden-mdb: What does "the parent is a single resource" mean in this context?
We would expect to see `listGroupsOperations` for `/api/atlas/v2/groups/operations`

tools/spectral/ipa/rulesets/IPA-104.yaml
julius-jogela: Fixed: reworded to describe what the rule expects instead of restating uniqueness.

tools/spectral/ipa/__tests__/utils/operationIdGeneration.test.js
julius-jogela: Fixed: removed redundant tests.

tools/spectral/ipa/__tests__/utils/operationIdGeneration.test.js
julius-jogela: Fixed: added a test for /groups/operations (no ID) — confirms it hits the collection-scoped branch and returns listGroupsOperations as expected. Also reworded the adjacent comment to name {groupId} as thats what makes the other case instance-scoped.

yelizhenden-mdb (approved):
LGTM, thanks!

PR #1434, opened 2026-08-12
feat: tag long-running operations with x-xgen-long-running-operation

julius-jogela:
## Proposed changes

This PR introduces automatic tagging of long-running operations with the `x-xgen-long-running-operation` extension.

Every operation that returns HTTP 202 is now tagged during the merge step, detected from the OpenAPI contract itself rather than tagged by hand by service teams. Operations that predate IPA-132 are tagged as legacy so that downstream tooling can tell them apart from operations that follow the long-running operation contract, instead of driving the same polling logic for both.

_Jira ticket:_ CLOUDP-429807

### Emitted shape

```yaml
# operation that follows the IPA-132 contract
x-xgen-long-running-operation:
  legacy: false

# operation that returns 202 but predates IPA-132
x-xgen-long-running-operation:
  legacy: true
```

Operations that do not return 202 carry no extension. The value is an object rather than a bare boolean so more fields can be added later without a breaking change.

### Where it is applied

Tagging happens in `foascli merge`, after the specs are federated. That output is published directly as `openapi/.raw/v2.yaml`, and `foascli filter` reads the same output to produce `openapi/v2.yaml`, so a single hook covers both. The extension also reaches the per-version specs under `openapi/v2`.

Downstream consumers such as the DevOps Integrations team read `openapi/v2.yaml`, so the extension had to survive filtering. No filter changes were needed.

### Identifying legacy operations

The spec carries no machine-readable signal that separates legacy 202 operations from compliant ones: The 17 legacy operations are therefore identified by a hardcoded list of operation IDs, isolated in one file behind `isLegacyLongRunningOperation`.

All 17 operations returning 202 today are legacy, so the list is currently also the complete set of 202 operations. Operations added later default to compliant, and are expected to follow the IPA-132 spectral rules.

### Changes

- `tools/cli/internal/cli/merge/long_running_operation.go` (new): the extension name, the published value type, the legacy operation list, and the tagger.
- `tools/cli/internal/cli/merge/merge.go`: one call to `tagLongRunningOperations` after the merge.
- `tools/cli/internal/cli/merge/long_running_operation_test.go` (new): compliant tagging, legacy tagging, absence of the tag on non-202 operations, all HTTP methods, unrelated extensions left untouched, the emitted JSON shape, nil safety, and the extension in the files the merge command writes.
- `tools/foas/openapi/filter/filter_test.go`: a `TestExtensionFiltering` case covering that the extension survives the filters that produce `openapi/v2.yaml` and the per-version specs, while `x-xgen-owner-team` and `x-xgen-IPA-exception` are still removed.

### Testing

All tests pass:

- Full e2e suite (`make e2e-test`): `ok`, 200s.
- Unit tests: 11 packages in `tools/cli`, 8 packages in `tools/foas`.
- `golangci-lint` and `gofmt`: clean.

The tagger was also run against the current `openapi/.raw/v2.json`: of 576 operations, the 17 that return 202 were tagged `legacy: true` and nothing else was tagged. Renaming one of them to an ID outside the list produced `legacy: false`, confirming the path for future compliant operations.

Adding the extension produces no changelog or breaking-change entries. Every oasdiff checker that reads an extension diff keys off a specific named extension (`x-sunset`, `x-extensible-enum`, `x-stability-level`), and this repo's changelog code never inspects extension diffs. Confirmed by running `changelog create` against revisions differing only by this extension: `changelog.json`, `changelog-all.json` and all `version-diff` files were byte-identical.


tools/cli/internal/cli/merge/merge.go
```diff
@@ -42,6 +42,8 @@ func (o *Opts) Run() error {
 		return err
 	}
 
+	tagLongRunningOperations(federated)
```
andmatei, line 45: I am wondering can if it would be a good idea to create a filter for this and add it [here](https://github.com/mongodb/openapi/tree/main/tools/foas/openapi/filter)?

tools/cli/internal/cli/merge/long_running_operation.go
```diff
@@ -0,0 +1,91 @@
...
+// limitations under the License.
+
+package merge
+
+import (
+	"github.com/mongodb/openapi/tools/foas/openapi"
+)
+
+const (
+	longRunningOperationExtension = "x-xgen-long-running-operation"
```
andmatei, line 22: [nit] I remember discussing if it would be a good idea to have `x-xgen-lro` instead so it is not as verbose, wondering where we landed on that

julius-jogela: I think we agreed on the fact that LRO is not an intuitive acronym and theres no reason to not be explicitly clear by using x-xgen-long-running-operation instead.

tools/cli/internal/cli/merge/long_running_operation_test.go
```diff
@@ -0,0 +1,209 @@
...
+		want        any
+		name        string
+		operationID string
+		statusCodes []string
+	}{
+		{
+			name:        "Compliant202IsTaggedAsNonLegacy",
+			operationID: "createGroupThing",
+			statusCodes: []string{"202"},
+			want:        longRunningOperation{Legacy: false},
```
yelizhenden-mdb, line 56: Is it feasible to omit this field for non-legacy endpoints and only set `legacy: true` when applicable? Wondering if that would satisfy both our internal workflows and downstream tooling, while keeping the OpenAPI spec lean

tools/cli/internal/cli/merge/merge.go
julius-jogela: A filter would tag the published spec but leave the raw one untagged, in split_spec.sh the merge output is moved straight to .raw/, and the filter commands only read it to produce v2.yaml. We want the extension in both.

tools/cli/internal/cli/merge/long_running_operation_test.go
julius-jogela: Its very feasible, just a readability tradeoff. Happy to go whichever direction you prefer.

yelizhenden-mdb: Could you clarify the readability tradeoff a bit?

julius-jogela: Sure                                                                                                           
                                                                                                                                                                                                              
 # current                                                                                                                                                                                                    
x-xgen-long-running-operation:                                                                                                                                                                               
legacy: false                                                                                                                                                                                                                                                                                                                                                                                                            

 # omitted                                                                                                                                                                                                                                                                                                                                                                          
x-xgen-long-running-operation: {}                                                                                                                                                                            
                                                                                                                                                                                                            
Both parse identically (absent can be read as false), so it's purely which one is nicer to read in v2.yaml. {} is more compact but might look like something got cut off.

yelizhenden-mdb: No strong opinion, thanks for double-checking. You can proceed with this approach

tools/foas/openapi/filter/long_running_operation_test.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/mongodb/openapi/tools/foas/apiversion"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// longRunningOperationExtension is set by the merge command and must pass through the filters
```
yelizhenden-mdb: Instead of creating a new test file, can you update `filter_test.go` with extension filtering cases? This way, we would preserve the repo structure?

julius-jogela: Done, moved the cases into filter_test.go

tools/cli/internal/cli/merge/long_running_operation_test.go
julius-jogela: Sticking with legacy: false. Thanks for taking a look

yelizhenden-mdb (approved):
LGTM

tools/cli/internal/cli/merge/merge.go
andmatei: Oh I see thank you so much for the clarification!!

andmatei (approved):
LGTM

PR #1435, opened 2026-08-13
feat(ipa): implement IPA-132 ruleset and base Operations resource rules

julius-jogela:
## Proposed changes

This PR adds the IPA-132 (Long-Running Operations) spectral ruleset with the three rules governing the shape of the Operations resource. It is the first of a series implementing [IPA-132](https://mongodb.github.io/ipa/132) validation.

_Jira ticket:_ CLOUDP-434369

### Rules added

All at `warn` severity while IPA-132 is experimental, with exceptions collected at the path-item level:

| Rule | Guideline | Checks |
| --- | --- | --- |
| `xgen-IPA-132-operations-endpoint-must-not-be-global` | [`IPA-132-operations-endpoint-must-not-be-global`](https://mongodb.github.io/ipa/132) | An Operations endpoint must have a parent resource in its path; a root-level `/api/atlas/v2/operations` is a violation |
| `xgen-IPA-132-operation-must-be-a-read-only-resource` | [`IPA-132-operation-must-be-a-read-only-resource`](https://mongodb.github.io/ipa/132) | Operations path items may define only `get` |
| `xgen-IPA-132-operations-endpoint-must-be-a-leaf-resource` | [`IPA-132-operations-endpoint-must-be-a-leaf-resource`](https://mongodb.github.io/ipa/132) | Nothing may be nested below `operations/{operationId}` |

The shared path predicates (identifying Operations endpoints in their collection-scoped, instance-scoped and `unauth` forms) live in the new `functions/utils/longRunningOperations.js`.

### Testing

- 27 new tests: per rule — valid shapes, violations (exact code/message/path/severity), violation-with-exception suppressed, adoption-with-exception flagged as unnecessary; plus unit tests for the path predicates.
- Full suite: 119 suites / 733 tests pass. `gen-ipa-docs`, prettier and eslint clean.
- `spectral lint openapi/.raw/v2.yaml` with the full ruleset: 0 `xgen-IPA-132-*` findings — the spec has no Operations endpoints yet.
- Live-fire check on a scratch spec with three planted violations (root-level `/operations`, `post` on an Operations endpoint, `/operations/{operationId}/logs`): each rule fires once at the right location.

## Checklist

### Changes to Spectral


tools/spectral/ipa/rulesets/IPA-132.yaml
```diff
@@ -0,0 +1,75 @@
...
+        - Applies to Operations endpoints, i.e. paths ending with an `operations` segment or an
+          `operations/{operationId}` suffix
+        - The `operations` segment must be preceded by at least one parent resource segment
+        - A root-level Operations endpoint, such as `/api/atlas/v2/operations` or
+          `/api/atlas/v2/unauth/operations`, is a violation
+        - Custom method suffixes (`:customMethod`) are ignored when identifying Operations endpoints
+        - Paths with `x-xgen-IPA-exception` for this rule are excluded from validation
+
+    message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-132-operations-endpoint-must-not-be-global'
+    severity: error
```
yelizhenden-mdb: We will start with `warn` severity as they are experimental at the moment

tools/spectral/ipa/rulesets/IPA-132.yaml
```diff
@@ -0,0 +1,75 @@
...
+    description: |
+      Operations endpoints are read-only and may only define the get method.
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+
+        - Applies to Operations endpoints, i.e. paths ending with an `operations` segment or an
+          `operations/{operationId}` suffix
+        - The path item must not define `post`, `put`, `patch` or `delete` methods
+        - Custom method paths attached to an Operations endpoint, such as
```
yelizhenden-mdb: We are not allowing any custom methods on Operations resources at the moment, and control actions like `cancel` would definitely be defined with a POST method, as they result in a side effect.

For now, my suggestion is to add a guideline to avoid custom methods on Operations endpoints, and to remove the custom method mentions from the rules.

tools/spectral/ipa/rulesets/functions/IPA132OperationsEndpointMustNotBeGlobal.js
```diff
@@ -0,0 +1,36 @@
...
+    return;
+  }
+
+  const errors = checkViolationsAndReturnErrors(input, path, ruleName);
+  return evaluateAndCollectAdoptionStatus(errors, ruleName, oas.paths[input], path);
+};
+
+function checkViolationsAndReturnErrors(input, path, ruleName) {
+  try {
+    if (isRootLevelOperationsPath(input)) {
```
yelizhenden-mdb: You can re-use `resourceBelongsToSingleParent` util function instead of defining a new `isRootLevelOperationsPath` function

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
```diff
@@ -0,0 +1,43 @@
...
+
+  const pathItem = oas.paths[input];
+  const errors = checkViolationsAndReturnErrors(pathItem, path, ruleName);
+  return evaluateAndCollectAdoptionStatus(errors, ruleName, pathItem, path);
+};
+
+function checkViolationsAndReturnErrors(pathItem, path, ruleName) {
+  try {
+    const errors = [];
+    for (const method of FORBIDDEN_METHODS) {
```
yelizhenden-mdb: You can re-use `isReadOnlyResource` util function

yelizhenden-mdb: JFYI `isReadOnlyResource` does not check the existence of other methods, it only looks at the schema fields. Existence of other methods is validated by other validation rules, such as `IPA-xxx-ReadOnlyResourceShouldNotHave(create/update/delete)Method` rules.

tools/spectral/ipa/rulesets/IPA-132.yaml
julius-jogela: Done in 9eb5a08cac — all three rules are now `warn`.

tools/spectral/ipa/rulesets/IPA-132.yaml
julius-jogela: Removed the custom method handling and mentions from the predicates, descriptions and tests in 9eb5a08cac. Agreed on the dedicated guideline — I'll add it to the IPA repo and follow up with the corresponding rule in the next PR once the guideline ID exists.

tools/spectral/ipa/rulesets/functions/IPA132OperationsEndpointMustNotBeGlobal.js
julius-jogela: I looked at reusing it, but `resourceBelongsToSingleParent` checks that the segment before the resource is a path parameter, so it returns `false` for the collection-scoped form `/groups/{groupId}/clusters/operations` (parent segment is `clusters`), and for `/parent/operations/{operationId}` it strips the trailing `{operationId}` plus one more segment and rejects that too — it would flag both shapes IPA-132 allows. It's also module-private in `resourceEvaluation.js`. I've added a note to the function doc explaining why it isn't reused; happy to export and generalize it instead if you'd prefer.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Right — and one more gap on top of that: the `xgen-IPA-106/107/108-readonly-resource-should-not-have-*-method` rules never evaluate the collection-scoped Operations form at all, because `isResourceCollectionIdentifier('/…/clusters/operations')` is `false` (the `/resource/resource` shape isn't a valid collection identifier). They'd also depend on every `OperationResponse` field being `readOnly: true`, which isn't guaranteed. This rule matches on the Operations path shape directly, so both forms are covered regardless of schema annotations.

tools/spectral/ipa/rulesets/functions/IPA132OperationsEndpointMustNotBeGlobal.js
yelizhenden-mdb: Thanks for double-checking! Exporting and generalizing it to check if a resource is root-level (i.e. has no parents) in the `resourceEvaluation` util file sounds like a good idea.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
yelizhenden-mdb: Good catch! We might need to revisit those validation rules.

For now, using `isReadOnlyResource` together with the `forbidden methods` looks like a good idea, with the forbidden methods passed as `functionOptions`.

OperationResource fields being `readOnly:true` must be guaranteed

tools/spectral/ipa/rulesets/functions/IPA132OperationsEndpointMustNotBeGlobal.js
julius-jogela: Done in 07d5ad070c - added an exported isRootLevelResource to resourceEvaluation.js and the rule uses that now. Left resourceBelongsToSingleParent as is since isSingletonResource depends on its current behaviour.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Done in 07d5ad070c - forbidden methods are passed as functionOptions now, and the rule also validates the Operation resource with isReadOnlyResource, so every property of the GET response schema has to be readOnly: true. The schema check runs once per resource (on the collection path item) so the same schema issue doesn't get reported on both paths. One thing to keep in mind for when we revisit the readonly-resource rules: for the collection-scoped form isReadOnlyResource only finds the GET through its fallback loop, since isSingleResourceIdentifier doesn't recognize /resource/operations/{operationId}. Works fine today, just leaning on the fallback.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const pathItem = oas.paths[input];
+  const errors = checkViolationsAndReturnErrors(input, pathItem, oas, forbiddenMethods, path, ruleName);
+  return evaluateAndCollectAdoptionStatus(errors, ruleName, pathItem, path);
+};
+
+function checkViolationsAndReturnErrors(input, pathItem, oas, forbiddenMethods, path, ruleName) {
+  try {
+    const errors = [];
+    for (const method of forbiddenMethods) {
```
yelizhenden-mdb: Taking a second look, I think we don't need the `forbidden methods`. If there are any methods defined other than Get or List (basically anything other than the GET verb), we can just say that it is not allowed. How does it sound?

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
```diff
@@ -0,0 +1,51 @@
...
+        errors.push({
+          path: [...path, method],
+          message: `Operations endpoints are read-only and do not allow the ${method} method.`,
+        });
+      }
+    }
+
+    // The readOnly schema condition applies to the resource as a whole, so it is evaluated once
+    // per Operations resource, on its collection path item
+    if (isOperationsCollectionPath(input) && !isReadOnlyResource(getResourcePathItems(input, oas.paths))) {
```
yelizhenden-mdb: >  The readOnly schema condition applies to the resource as a whole, so it is evaluated once per Operations resource, on its collection path item

The best place to validate that the resource is read-only is the instance path, as that is where the Get method is defined. The collection path would be the List method, which already uses the Operation resource schema.

I think the workflow can go like this:
1. Ensure it is an Operations resource
2. Check whether the path has any methods defined other than `get` (if the number of defined methods is more than one, it probably has methods other than get; otherwise, if the length is 1, check that it is get)
3. If the path is an Operation instance path, check that it is a read-only resource

How does it sound?

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Good call, done - dropped the functionOptions entirely, the rule now flags any HTTP method other than get on an Operations endpoint. While in there I also moved the readOnly schema check to run per path item on the get method's 2xx schemas (still via allPropertiesAreReadOnly): the resource-level isReadOnlyResource call turned out to depend on spec key order for the collection-scoped form and couldn't be suppressed by an exception on the path item that actually holds the offending get.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Sounds right, done in 1a4a51d653 - dropped the functionOptions entirely, the rule now flags any HTTP method other than get on an Operations endpoint (put, post, delete, patch, head, options, trace).

tools/spectral/ipa/__tests__/IPA132OperationsEndpointMustNotBeGlobal.test.js
```diff
@@ -0,0 +1,122 @@
...
+      },
+      {
+        code: 'xgen-IPA-132-operations-endpoint-must-not-be-global',
+        message: ERROR_MESSAGE,
+        path: ['paths', '/api/atlas/v2/unauth/operations'],
+        severity: DiagnosticSeverity.Warning,
+      },
+    ],
+  },
+  {
```
yelizhenden-mdb: I haven't seen such a case before in OpenAPI specs, we can remove this one

tools/spectral/ipa/__tests__/IPA132OperationsEndpointMustNotBeGlobal.test.js
```diff
@@ -0,0 +1,122 @@
...
+          'x-xgen-IPA-exception': {
+            'xgen-IPA-132-operations-endpoint-must-not-be-global': 'reason',
+          },
+        },
+      },
+    },
+    errors: [],
+  },
+  {
+    name: 'nested Operations endpoints do not need an exception',
```
yelizhenden-mdb: This is not about validating the globalness of Operation resource, we can remove this test case

tools/spectral/ipa/__tests__/IPA132OperationMustBeAReadOnlyResource.test.js
```diff
@@ -0,0 +1,215 @@
...
+  },
+  {
+    name: 'invalid Operations endpoints with methods other than get',
+    document: {
+      paths: {
+        '/api/atlas/v2/resourceName/operations': {
+          get: readOnlyGet,
+          post: {},
+        },
+        '/api/atlas/v2/resourceName/operations/{operationId}': {
```
yelizhenden-mdb, line 81: Can we move the examples to the top of the file and reuse them in the test cases?

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Makes sense, done in 67b8faf6d0 - the readOnly check now runs on the single Operation endpoint only, where the Get method is defined. The method check still runs on every operations path item (anything other than get gets flagged).

tools/spectral/ipa/__tests__/IPA132OperationsEndpointMustNotBeGlobal.test.js
julius-jogela: Removed in 67b8faf6d0.

tools/spectral/ipa/__tests__/IPA132OperationsEndpointMustNotBeGlobal.test.js
julius-jogela: Removed in 67b8faf6d0.

tools/spectral/ipa/__tests__/IPA132OperationMustBeAReadOnlyResource.test.js
julius-jogela: Done in 67b8faf6d0 - the shared path item examples are defined once at the top of the file and reused across the test cases.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const pathItem = oas.paths[input] ?? {};
+  const errors = checkViolationsAndReturnErrors(input, pathItem, path, ruleName);
+  return evaluateAndCollectAdoptionStatus(errors, ruleName, pathItem, path);
+};
+
+function checkViolationsAndReturnErrors(input, pathItem, path, ruleName) {
+  try {
+    const errors = [];
+    for (const method of HTTP_METHODS) {
```
yelizhenden-mdb: So we are currently checking whether all the methods are present for the path, am I correct?

Instead, we can follow the same approach as the `IPA113ResetMethodMustUsePost` validation rule implementation: in that case, it checks if it is only post; for our case, it would be get.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
```diff
@@ -0,0 +1,68 @@
...
+        errors.push({
+          path: [...path, method],
+          message: `Operations endpoints are read-only and do not allow the ${method} method.`,
+        });
+      }
+    }
+
+    // The readOnly condition is checked on the single Operation endpoint, where the Get method is
+    // defined. The List method on the collection reuses the Operation resource schema and its
+    // response shape is validated by xgen-IPA-132-operation-endpoints-must-return-operation-response.
```
yelizhenden-mdb: Do we have `xgen-IPA-132-operation-endpoints-must-return-operation-response` rule in place?

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+export default (input, _, { path, documentInventory, rule }) => {
+  const ruleName = rule.name;
+  const oas = documentInventory.resolved;
+
+  if (!isOperationsPath(input)) {
+    return;
+  }
+
+  const pathItem = oas.paths[input] ?? {};
```
yelizhenden-mdb: Could you clarify why we have this line of code?

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
```diff
@@ -0,0 +1,68 @@
...
+    if (isSingleOperationPath(input) && hasWritableGetResponseSchema(pathItem)) {
+      errors.push({ path: [...path, 'get'], message: READ_ONLY_SCHEMA_ERROR_MESSAGE });
+    }
+    return errors;
+  } catch (e) {
+    return handleInternalError(ruleName, path, e);
+  }
+}
+
+function hasWritableGetResponseSchema(pathItem) {
```
yelizhenden-mdb: [nit] How about `hasReadOnlyGetResponseSchema` and negating it? `WritableGetResponseSchema` sounds a bit awkward.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
yelizhenden-mdb: Could you please create a ticket to note these limitations so we can work on them later?

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Done in 58803309b3 - same approach as IPA113ResetMethodMustUsePost, the defined method keys are extracted from the path item and anything other than get is flagged.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Not in this PR - it's part of the follow-up (#1437, stacked on this one) that adds the OperationResponse schema rules. Removed the forward reference from the comment in 58803309b3 so this PR stands on its own.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: It was a fallback for null path items, but since that's not a case we see in practice I removed it in 58803309b3.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Done in 58803309b3.

tools/spectral/ipa/rulesets/functions/IPA132OperationMustBeAReadOnlyResource.js
julius-jogela: Will do - creating a ticket to track the resource evaluation util limitations (the collection-scoped Operations path gaps and the isReadOnlyResource fallback behaviour), will link it here.

julius-jogela: https://jira.mongodb.org/browse/CLOUDP-435907

yelizhenden-mdb (approved):

PR #1437, opened 2026-08-14
feat(ipa): add IPA-132 OperationResponse schema rules

julius-jogela:
## Proposed changes

This PR adds the IPA-132 rules validating how long-running operation status is communicated through the `OperationResponse` schema. The schema checks validate the [OperationResponse reference schema](https://mongodb.github.io/ipa/132#operationresponse-schema) by check type, with the field lists declared in the ruleset's `functionOptions`.

_Jira ticket:_ CLOUDP-429837

### Rules added

All at `warn` severity, anchored on `$.components.schemas.OperationResponse` (shared `OperationResponseSchema` alias):

| Rule | Checks |
| --- | --- |
| `xgen-IPA-132-operation-endpoints-must-return-operation-response` | Every 2xx response of a GET on an Operations endpoint must carry a JSON schema; the single Operation endpoint (`.../operations/{operationId}`) must reference the `OperationResponse` schema; the Operations collection endpoint must reference a paginated wrapper defining the `results` array directly (inline and `allOf`-composed wrappers rejected, consistent with the IPA-110 pagination rules). Responses defined as `$ref`s to `components.responses` are followed. Exceptions at the response level |
| `xgen-IPA-132-operation-response-must-define-required-fields` | The schema is flat (no composition) and defines `operationId`, `status`, `operationType`, `createdAt`, `updatedAt`, `expiresAt`, each listed as required |
| `xgen-IPA-132-operation-response-must-define-optional-fields` | `customMethod`, `error`, `resultHref`, `retryAfterSeconds` defined and **not** listed as required — each is present only under a runtime condition that cannot be validated statically |
| `xgen-IPA-132-operation-response-must-use-standard-enum-values` | `status` enum exactly `PENDING, IN_PROGRESS, SUCCEEDED, FAILED, CANCELED, SUPERSEDED`; `operationType` enum exactly `CREATE, UPDATE, DELETE, CUSTOM`; `error.retryStrategy` enum exactly `IMMEDIATE, BACKOFF, NONE` |
| `xgen-IPA-132-operation-response-must-define-nested-object-structure` | The `error` object must define `code`, `message` and `retryable` |
| `xgen-IPA-132-operations-should-expose-status-fields` | The schema should define `statusMessage`, `progress` (with `completed`, `total`, `unit`) and `estimatedCompletionTime`. Kept separate because it enforces should-level guidance and severity is per-rule |

Each check-type rule skips fields whose presence another rule owns, so a missing field is reported exactly once. Each rule's description lists the IPA-132 guidelines it covers.

There is one and only one `OperationResponse`, provided as scaffolding, so the schema must define its properties directly: `allOf`/`oneOf`/`anyOf` composition is a violation reported by the required-fields rule, and the remaining schema rules skip composed schemas.

### Testing

- Per rule: valid, invalid (exact error count, message and path), violation-with-exception (suppressed) and adoption-with-exception (flagged as unnecessary) cases, plus unit tests for the composition and enum helpers.
- Full suite: 124 suites / 786 tests pass. `gen-ipa-docs`, prettier and eslint clean.
- `spectral lint openapi/.raw/v2.yaml` with the full ruleset: 0 `xgen-IPA-132-*` findings — the spec defines no `OperationResponse` schema yet.

## Checklist


tools/spectral/ipa/__tests__/IPA132OperationEndpointsMustReturnOperationResponse.test.js
```diff
@@ -0,0 +1,452 @@
...
+      },
+    },
+    PaginatedMetadata: {
+      type: 'object',
+      properties: {
+        totalCount: { type: 'integer' },
+      },
+    },
+    // A paginated wrapper composed with allOf, which Spectral does not flatten
+    PaginatedAllOfOperationResponse: {
```
yelizhenden-mdb, line 50: Why do we allow `allOf`? Could you clarify that?

tools/spectral/ipa/__tests__/IPA132OperationResponseMustDefineRequiredFields.test.js
```diff
@@ -0,0 +1,213 @@
...
+const CUSTOM_METHOD_MISSING_ERROR_MESSAGE =
+  'OperationResponse must define a customMethod property for custom method operations.';
+const CUSTOM_METHOD_REQUIRED_ERROR_MESSAGE =
+  'The customMethod property must not be listed as required. It is required only when operationType is CUSTOM.';
+
+const validOperationResponse = {
+  type: 'object',
+  required: ['operationId', 'operationType', 'createdAt', 'updatedAt'],
+  properties: {
+    operationId: { type: 'string' },
```
yelizhenden-mdb, line 8: Looking at the [Technical Design document](https://docs.google.com/document/d/1QZLurLgmHC4oTPoG1UZQZntX2oOFLtzdhKc8S0isYYo/edit?tab=t.tpl9rs5lgose#bookmark=id.xzbmw6693j00), I see that there are other `required` fields, in other words, core metadata. Shouldn't they be involved here as well?

tools/spectral/ipa/rulesets/IPA-132.yaml
```diff
@@ -72,3 +75,72 @@ rules:
...
+          defined and listed as required
+        - The `operationType` enum must contain exactly the values `CREATE`, `UPDATE`, `DELETE`
+          and `CUSTOM`, in any order
+        - A `customMethod` property must be defined and must not be listed as required: it is
+          required only when `operationType` is `CUSTOM`, which cannot be validated statically
+        - Schemas with `x-xgen-IPA-exception` for this rule are excluded from validation
+
+    message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-132-operation-response-must-include-core-metadata'
+    severity: warn
+    given: $.components.schemas.OperationResponse
```
yelizhenden-mdb: Can you define an alias for it and re-use later?

tools/spectral/ipa/__tests__/IPA132OperationEndpointsMustReturnOperationResponse.test.js
julius-jogela: Spectral resolves `$ref`s but does not flatten `allOf`, so without merging the `allOf` branches a composed-but-compliant wrapper would be a false violation — per OAS semantics an instance must satisfy every subschema, so a `results` array defined in any branch is genuinely part of the schema. Root-level `allOf` composition is a common idiom in our spec (107 of the 1820 schemas in `openapi/.raw/v2.yaml` are composed this way), so the first real paginated Operations wrapper may well extend a shared pagination base like this. This test pins that case.

tools/spectral/ipa/__tests__/IPA132OperationResponseMustDefineRequiredFields.test.js
julius-jogela: Deliberate split, each rule maps 1:1 onto a guideline ID in the IPA repo, and [IPA-132-operation-response-must-include-core-metadata](https://mongodb.github.io/ipa/132) scopes core metadata to `operationId`, `operationType`, `createdAt` and `updatedAt` (plus `customMethod` declared but not required). The other required fields in the design doc are owned by their own guidelines:

- `status` → `xgen-IPA-132-operation-status-must-use-the-standard-status-enum` (this PR)
- `resultHref`, `retryAfterSeconds`, `error` → conditionally required at runtime, so statically they can only be checked as declared-and-not-required; covered by `xgen-IPA-132-successful-lros-must-reference-the-result-resource`, `xgen-IPA-132-in-progress-operations-must-expose-retry-after-seconds` and `xgen-IPA-132-failed-lros-must-report-a-structured-error` in the next PR.
- `expiresAt` → owned by `IPA-132-operations-must-expire-after-a-finite-retention-period`, which is review-level in the IPA repo because retention and the 404-after-expiry are backend behavior

tools/spectral/ipa/rulesets/IPA-132.yaml
julius-jogela: Done in d70c98d6c8 — added an `OperationResponseSchema` alias and pointed both schema rules at it. The Operation outcome/progress rules coming in the next PR will reuse it as well.

tools/spectral/ipa/__tests__/IPA132OperationResponseMustDefineRequiredFields.test.js
yelizhenden-mdb: I recall a discussion where we said that since there are several guidelines covering `required` fields, the plan was to cover all of them in one validation rule, so we don't need a separate validation rule for each required-field guideline. Isn't that still the case?

tools/spectral/ipa/__tests__/IPA132OperationEndpointsMustReturnOperationResponse.test.js
yelizhenden-mdb: My question is more about whether we expect allOf usage in the Operations endpoints

julius-jogela: Update on the allOf case after a deeper review pass: I've kept allOf support for the `OperationResponse` schema itself (now merged conjunctively — re-declared properties combine keywords and enums intersect), but dropped allOf acceptance for the *paginated wrapper*. The wrapper shape I blessed here would fail `xgen-IPA-110-collections-response-define-results-array` at error severity on `/…/{id}/operations`, since `schemaIsPaginated` doesn't traverse allOf — so IPA-132 now requires the `results` array directly on the wrapper, consistent with IPA-110. Teaching the IPA-110 pagination predicate allOf would change an existing error-level rule and belongs in its own PR.

tools/spectral/ipa/__tests__/IPA132OperationResponseMustDefineRequiredFields.test.js
julius-jogela: Restructured in 907d1c6b57 following our Slack discussion. The schema checks are now split by check type instead of by guideline, with the field lists in functionOptions: one rule for all the always-required fields (operationId, status, operationType, createdAt, updatedAt), one for the optional/conditional ones, one for enum values and one for nested structure.

Two things I kept on purpose: operations-should-expose-status-fields stays its own rule since severity is per-rule and that one is a should, and expiresAt still isn't linted because the expiry guideline is review-level. If we decide it should be required, it's a guideline change in the IPA repo first and then a one-line functionOptions addition here.

tools/spectral/ipa/rulesets/IPA-132.yaml
```diff
@@ -72,3 +82,195 @@ rules:
...
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+
+        - Applies to the OperationResponse component schema
+        - Each field listed in the rule's functionOptions must be defined and listed as required
+        - Properties and required lists spread across allOf sub-schemas are merged before
+          validation; composition through oneOf/anyOf is not merged and is reported as missing
+          fields
+        - Covers the required-field checks of the IPA-132-operation-status-must-use-the-standard-status-enum,
```
yelizhenden-mdb: We do not need to mention guideline-rule mappings. GuidelineIDs are internal to MongoDB, not visible on the documentation page

tools/spectral/ipa/rulesets/IPA-132.yaml
```diff
@@ -72,3 +82,195 @@ rules:
...
+    description: |
+      The OperationResponse schema must define every field the IPA-132 standard marks as always
+      required: operationId, status, operationType, createdAt, updatedAt and expiresAt.
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+
+        - Applies to the OperationResponse component schema
+        - Each field listed in the rule's functionOptions must be defined and listed as required
+        - Properties and required lists spread across allOf sub-schemas are merged before
```
yelizhenden-mdb: Could you clarify allOf/oneOf/anyOf references and how they are related to OperationResponse schema?

tools/spectral/ipa/rulesets/IPA-132.yaml
```diff
@@ -72,3 +82,195 @@ rules:
...
+      enum of the error object.
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+
+        - Applies to the OperationResponse component schema
+        - Each enum field listed in the rule's functionOptions must use exactly the listed values,
+          in any order
+        - Fields that are not defined are skipped: top-level field presence is validated by the
+          required-fields and optional-fields rules, and the nested `retryStrategy` field is
```
yelizhenden-mdb:
```suggestion
          required-fields and optional-fields rules
```

tools/spectral/ipa/rulesets/IPA-132.yaml
julius-jogela: Removed in d498106081 — dropped the guideline-ID mappings from all four rule descriptions. Kept the one user-facing part without IDs: that for expiresAt only the field declaration is validated, the retention behavior is backend-defined.

tools/spectral/ipa/rulesets/IPA-132.yaml
julius-jogela: Reworded in d498106081. There are cases where the schema comes out allOf-composed — the generator emits Java inheritance as allOf (e.g. AWSCloudProviderContainer) — and we want to catch these and validate them correctly. oneOf/anyOf describe alternatives, so they aren't merged and get flagged.

tools/spectral/ipa/rulesets/IPA-132.yaml
julius-jogela: Applied in d498106081.

tools/spectral/ipa/rulesets/IPA-132.yaml
julius-jogela: Follow-up: after confirming the design (one and only one OperationResponse, provided as scaffolding), I dropped the allOf merge entirely in 7cda85d6f4. The schema must now define its properties directly — allOf/oneOf/anyOf composition is an explicit violation reported by the required-fields rule, and the other schema rules skip composed schemas. Much less machinery, and it matches the design instead of hedging against it.

tools/spectral/ipa/__tests__/IPA132OperationEndpointsMustReturnOperationResponse.test.js
julius-jogela: Final state after our Slack discussion: composition is not expected anywhere for the Operation status contract, so as of 7cda85d6f4 the rules enforce that directly — OperationResponse must be a flat schema (allOf/oneOf/anyOf is a violation), and the paginated wrapper must define results directly per IPA-110. The allOf merge machinery is removed.

tools/spectral/ipa/rulesets/functions/IPA132OperationEndpointsMustReturnOperationResponse.js
```diff
@@ -0,0 +1,116 @@
...
+  isOperationsCollectionPath,
+  isSingleOperationPath,
+  OPERATION_RESPONSE_SCHEMA_NAME,
+} from './utils/longRunningOperations.js';
+
+const SINGLE_OPERATION_ERROR_MESSAGE = `The Operation endpoint must return the ${OPERATION_RESPONSE_SCHEMA_NAME} schema.`;
+const OPERATIONS_COLLECTION_ERROR_MESSAGE = `The Operations collection endpoint must return a paginated response whose results reference the ${OPERATION_RESPONSE_SCHEMA_NAME} schema.`;
+const MISSING_SCHEMA_ERROR_MESSAGE = 'The Operations endpoint must define a JSON response schema.';
+
+const COMPONENT_RESPONSES_REF_PREFIX = '#/components/responses/';
```
yelizhenden-mdb: In our spec `components/responses` is reserved for error responses. Every 2xx response defines content inline or references the schema via `components/schemas`. Since this rule only evaluates 2xx codes, `resolveResponseRef` and the `contentBasePath` branch are unreachable. Could we drop them and read `response.content` directly?

tools/spectral/ipa/rulesets/functions/IPA132OperationResponseMustDefineRequiredFields.js
```diff
@@ -0,0 +1,48 @@
+import { evaluateAndCollectAdoptionStatus, handleInternalError } from './utils/collectionUtils.js';
+import { usesComposition } from './utils/longRunningOperations.js';
+
+const COMPOSITION_ERROR_MESSAGE =
```
yelizhenden-mdb: `IPA132OperationResponseMustDefineRequiredFields` owns two unrelated checks: "don't use allOf/oneOf/anyOf" and "define these six fields." Because they share a rule name, they share an exception as well, and that opens a gap.

If you ran a schema composed via `allOf` and missing `operationId` entirely through this validation, plus one exception:

```
"x-xgen-IPA-exception": {
  "xgen-IPA-132-operation-response-must-define-required-fields": "we compose from a base schema"
}
```

Result: 0 warnings. The exception suppresses the composition error.

Suggested fix: extract it into its own rule, `xgen-IPA-132-operation-response-must-not-use-composition`, on the same `OperationResponseSchema` alias. That matches the naming already used here, gives composition its own exception key and its own adoption metric, and leaves each rule named for what it checks.

Please update the other rules accordingly if needed

tools/spectral/ipa/rulesets/functions/IPA132OperationEndpointsMustReturnOperationResponse.js
julius-jogela: You're right — removed. `resolveResponseRef` and `contentBasePath` are gone and the rule reads `response.content` directly. Rule description, README, and tests updated accordingly.

tools/spectral/ipa/rulesets/functions/IPA132OperationResponseMustDefineRequiredFields.js
julius-jogela: Good catch, extracted as suggested. Composition rejection now lives in `xgen-IPA-132-operation-response-must-not-use-composition` with its own exception key and adoption metric. Your bypass scenario — an `allOf` schema with a required-fields exception — is now a test case and produces the composition violation.

yelizhenden-mdb (approved):
LGTM, thanks for addressing all the feedback

PR #1447, opened 2026-08-24
refactor(ipa): single-source the legacy long-running operation list

julius-jogela:
## Proposed changes

This PR moves the hardcoded list of 17 legacy long-running operationIds out of the foascli merge tagger into a single shared file, so the upcoming IPA-132 LRO rules can consume the same list the tagger uses. Tagging behavior is unchanged.

_Jira ticket:_ CLOUDP-438234

### Why

The IPA-132 LRO rules will select on the declared 202 response so they also work on the raw per-service specs in MMS, where `x-xgen-long-running-operation` does not exist (it is only added at the merge step). Those rules need the same legacy exclusions the tagger applies, without maintaining the list twice.

### How

- `tools/spectral/ipa/rulesets/functions/utils/legacyLongRunningOperations.json` — the single source of truth. Lives inside the ruleset package so it ships to npm consumers.
- The Go map in the merge package is now a checked-in file generated from that JSON (`go generate ./internal/cli/merge`, generator under `internal/cli/merge/gen`). `go:embed` cannot cross the `tools/cli` module boundary and the released foascli binary cannot read repo files at runtime, so generation is the mechanism.
- A new unit test reads the JSON and asserts set-equality with the generated map, so the two cannot drift; it runs in the existing unit-test CI step.

### Testing

- `go build ./...`, `go vet`, `go test -race -cover ./internal...` — all pass; the existing `tagLongRunningOperations` tests are unchanged and green.
- `golangci-lint run` (v2.10.1, the CI pin): 0 issues.
- `go generate` verified idempotent; regenerated output is byte-identical to the previous hand-written map.
- Repo-wide prettier check passes (covers the new JSON).

## Checklist


tools/cli/internal/cli/merge/gen/main.go
```diff
@@ -0,0 +1,96 @@
...
+
+	sort.Strings(list.OperationIDs)
+
+	var builder strings.Builder
+	builder.WriteString("// Code generated by go generate; DO NOT EDIT.\n")
+	builder.WriteString("// Source: tools/spectral/ipa/rulesets/functions/utils/legacyLongRunningOperations.json\n")
+	builder.WriteString("//\n")
+	builder.WriteString("// Generated by this command:\n")
+	builder.WriteString("//\n")
+	builder.WriteString("//\tgo run ./gen\n\n")
```
yelizhenden-mdb, line 75: Q: Does this command run automatically as part of the build, or do we need to run it manually?
(If manual, it would be better if we can add it to somewhere in the contributing.md)

julius-jogela: It's manual, go generate doesn't run as part of the build. There's a unit test that compares the generated map against the JSON so the two can't drift. Also added a line to CONTRIBUTING.md for clarity.

yelizhenden-mdb (approved):

PR #1448, opened 2026-08-24
feat(ipa): add IPA-132 long-running operation initiation rules

julius-jogela:
## Proposed changes

This PR adds the three IPA-132 rules validating how a long-running operation is initiated: 202 Accepted with a Location header carrying the operation handle, no other success status code, and no response body.

_Jira ticket:_ CLOUDP-435358

### Rules added

All at `warn` severity, given `$.paths[*][post,put,patch,delete]`:

| Rule | Checks |
| --- | --- |
| `xgen-IPA-132-lro-initiation-must-return-202-accepted-with-location` | The 202 response declares a `Location` header, using that exact casing |
| `xgen-IPA-132-lro-must-not-return-success-status-code-other-than-202` | No 2xx response other than 202 is declared, including wildcard codes such as `2XX` — blocks the "200 if fast, 202 if slow" pattern |
| `xgen-IPA-132-lro-202-response-must-not-include-content` | No media type of the 202 response declares a schema; media types carrying only versioning metadata (`x-xgen-version`) are allowed, consistent with the IPA-108 delete-response precedent |

### How long-running operations are selected

The rules select on the **declared 202 response**, through a single `isNonLegacyLongRunningOperation` helper. The `x-xgen-long-running-operation` extension is derived from the 202 during the merge step, so selecting on the 202 gives identical results on the federated spec — while also working on the raw per-service specs, where the extension does not exist and where the validation feedback reaches API authors. Legacy operations that predate IPA-132 are excluded, whether marked `legacy: true` by the merge step or matched by the shared legacy operationId list from #1447.

The shared list becomes a plain JavaScript module (`legacyLroOperationIds.js`) that the rules import directly, replacing #1447's JSON file: Spectral bundles rule functions with a rollup that cannot import JSON, so a data module is the only bundler-safe carrier. `go generate` extracts the Go copy from that module instead; the existing Go-side drift test keeps the two in sync.

### Testing

- Per rule: valid, invalid (exact error count, message and path), violation-with-exception (suppressed) and compliant-with-exception (flagged as unnecessary) cases; selection cases covering untagged operations declaring a 202 (violating and valid), operations tagged without a 202 (ignored), synchronous operations (ignored), and legacy exclusion via both the `legacy: true` marking and the shared operationId list. Unit tests for `isNonLegacyLongRunningOperation` and the shared-list wiring.
- Full suite: 132 suites / 846 tests pass. `gen-ipa-docs`, prettier and eslint clean.
- `spectral lint openapi/.raw/v2.yaml` with the full ruleset: 0 `xgen-IPA-132-*` findings — all 17 legacy 202 operations are excluded by the shared list on the untagged spec.
- Go side: `go build`, `go test -race`, golangci-lint (v2.10.1) clean; both generator outputs verified idempotent.

## Checklist


tools/spectral/ipa/rulesets/functions/utils/longRunningOperations.js
```diff
@@ -103,6 +111,29 @@ export function operationsSegmentIsLeaf(path) {
...
 
+export const LRO_EXTENSION = 'x-xgen-long-running-operation';
+
+/**
+ * Checks if an operation is a long-running operation that must follow the IPA-132 contract. A
+ * long-running operation declares a 202 response: the x-xgen-long-running-operation extension is
+ * derived from the 202 during the merge step, so selecting on the declared 202 gives the same
+ * result on the federated spec while also working on specs the merge step has not processed,
+ * such as the per-service source specs. Legacy operations predate IPA-132 and are excluded,
+ * whether marked `legacy: true` by the merge step or matched by the legacy operationId list.
```
yelizhenden-mdb: Q: Do we have `legacy:true` marking implemented? Or, is it planned?

tools/spectral/ipa/rulesets/functions/utils/longRunningOperations.js
```diff
@@ -103,6 +111,29 @@ export function operationsSegmentIsLeaf(path) {
...
+ *
+ * @param {object} operation the operation object to evaluate
+ * @returns {boolean} true if the operation is a long-running operation that is not legacy
+ */
+export function isCompliantLongRunningOperation(operation) {
```
yelizhenden-mdb: The naming seems slightly misleading. We don’t yet know whether these operations are compliant. As the description suggests, they are expected to follow the IPA-132 contract, but they haven’t been validated yet. Is that correct?

tools/spectral/ipa/rulesets/functions/IPA132LroInitiationMustReturn202AcceptedWithLocation.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  const errors = checkViolationsAndReturnErrors(input, path, ruleName);
+  return evaluateAndCollectAdoptionStatus(errors, ruleName, input, path);
+};
+
+function checkViolationsAndReturnErrors(operation, path, ruleName) {
+  try {
+    const headers = operation.responses['202'].headers;
+    // HTTP header names are case-insensitive
```
yelizhenden-mdb: I know that HTTP header names are case-insensitive, so Content-Type, content-type, and CONTENT-TYPE are treated the same. However, for consistency in our spec, we can require all headers to use the same casing convention as `Location`

tools/spectral/ipa/rulesets/functions/utils/longRunningOperations.js
julius-jogela: Yes, it's implemented and merged: `tagLongRunningOperations` in `tools/cli/internal/cli/merge/long_running_operation.go` sets `x-xgen-long-running-operation: {legacy: <bool>}` on every operation declaring a 202 (and strips it otherwise), with the flag derived from the shared operationId list single-sourced in #1447.

Worth noting the consequence for this helper: on the federated spec the extension check and the operationId-list check draw on the same source and are equivalent, so the list check is really what carries the per-service raw specs, where the merge step has not run.

tools/spectral/ipa/rulesets/functions/utils/longRunningOperations.js
julius-jogela: You are right, that reads as "already validated" when all it does is select what is in scope. Renamed to `isNonLegacyLongRunningOperation`, which states just what it tests — declares a 202 and is not legacy — and reworded the JSDoc to spell out that it says nothing about whether the contract is followed yet.

tools/spectral/ipa/rulesets/functions/IPA132LroInitiationMustReturn202AcceptedWithLocation.js
julius-jogela: Agreed, consistency in the spec beats HTTP's leniency here. The rule now requires the exact `Location` key, with a distinct message when a case variant is present (`The 202 Accepted response header must be named 'Location', found 'location'.`) so the feedback points at the casing rather than reading as a missing header.

No effect on the current spec: `openapi/v2.yaml` has no `Location` response header at all today, and the existing response headers already use canonical casing (`RateLimit-Limit`, `RateLimit-Remaining`).

yelizhenden-mdb (approved):

PR #1461, opened 2026-08-28
feat(ipa): require long-running operations to expose an Operations endpoint

julius-jogela:
## Proposed changes

This PR adds the last IPA-132 rule: a long-running operation must expose the Operations endpoint clients poll for status.

_Jira ticket:_ CLOUDP-435359

### Rule added

At `warn` severity, given `$.paths[*][post,put,patch,delete]`:

| Rule | Checks |
| --- | --- |
| `xgen-IPA-132-lros-must-expose-an-operations-endpoint` | Both the Operations collection (`.../operations`) and the single Operation endpoint (`.../operations/{operationId}`) are defined under the resource path of the long-running operation |

Resource path resolution: custom method suffixes are stripped before deriving the resource path. A mutation of a collection must expose the endpoints under the collection path (`/resource/operations`), and a mutation of a single resource under the resource instance path (`/resource/{resourceId}/operations`).

### How long-running operations are selected

Through the shared `isNonLegacyLongRunningOperation` helper introduced in #1448 — selection is on the declared 202 response, with legacy operations excluded both by the `legacy: true` marking and by the shared legacy operationId list. No new selection logic.

### Testing

Valid cases for the collection-level, instance-level and custom-method shapes; invalid cases for a missing Operations endpoint, for a collection defined without the single Operation endpoint, and for an instance-level operation whose Operations endpoints sit only at the parent collection; violation-with-exception (suppressed) and compliant-with-exception (flagged as unnecessary); selection cases covering untagged operations declaring a 202 (valid and violating), synchronous operations, operations tagged without a 202, and legacy exclusion via both the `legacy: true` marking and the shared operationId list.

Full suite: 131 suites / 861 tests pass. Prettier clean.

## Checklist

### DIFF
--- a/tools/spectral/ipa/rulesets/IPA-132.yaml
+++ b/tools/spectral/ipa/rulesets/IPA-132.yaml
@@ -18,6 +18,7 @@ functions:
   - IPA132LroInitiationMustReturn202AcceptedWithLocation
   - IPA132LroMustNotReturnSuccessStatusCodeOtherThan202
   - IPA132Lro202ResponseMustNotIncludeContent
+  - IPA132LrosMustExposeAnOperationsEndpoint
 
 aliases:
   OperationResponseSchema:
@@ -420,3 +421,34 @@ rules:
     given: $.paths[*][post,put,patch,delete]
     then:
       function: IPA132Lro202ResponseMustNotIncludeContent
+
+  xgen-IPA-132-lros-must-expose-an-operations-endpoint:
+    description: |
+      A long-running operation must expose an Operations endpoint through which clients poll the
+      operation status.
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+
+        - Applies to mutating methods that declare a 202 response; the
+          `x-xgen-long-running-operation` extension is derived from the 202 during the merge
+          step, so the rule behaves identically on merged and per-service specs
+        - Legacy operations that predate IPA-132 are excluded from validation, whether marked
+          `legacy: true` during the merge step or matched by the shared legacy operationId list
+        - Both the Operations collection (`.../operations`) and the single Operation endpoint
+          (`.../operations/{operationId}`) must be defined, nested under the resource path of the
+          long-running operation
+        - Each endpoint must define the `get` method, i.e. List on the Operations collection and
+          Get on the single Operation; the Operations resource is read-only, so `get` is the only
+          method through which clients can poll status
+        - A mutation of a collection must expose the endpoints under the collection path
+          (`/resource/operations`), and a mutation of a single resource under the resource
+          instance path (`/resource/{resourceId}/operations`)
+        - Custom method suffixes are ignored when deriving the resource path
+        - Operations with `x-xgen-IPA-exception` for this rule are excluded from validation
+
+    message: '{{error}} https://mdb.link/mongodb-atlas-openapi-validation#xgen-IPA-132-lros-must-expose-an-operations-endpoint'
+    severity: warn
+    given: $.paths[*][post,put,patch,delete]
+    then:
+      function: IPA132LrosMustExposeAnOperationsEndpoint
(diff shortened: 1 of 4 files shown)


tools/spectral/ipa/rulesets/functions/IPA132LrosMustExposeAnOperationsEndpoint.js
```diff
@@ -0,0 +1,63 @@
...
+  const errors = checkViolationsAndReturnErrors(path[1], oas, path, ruleName);
+  return evaluateAndCollectAdoptionStatus(errors, ruleName, input, path);
+};
+
+function checkViolationsAndReturnErrors(resourcePath, oas, path, ruleName) {
+  try {
+    const basePath = isCustomMethodIdentifier(resourcePath) ? stripCustomMethodName(resourcePath) : resourcePath;
+
+    const candidatePaths = [basePath];
+    if (isSingleResourceIdentifier(basePath)) {
```
yelizhenden-mdb: if the path is collection identifier -> check resource/operations and resource/operations/{operationId} exists
if the path is single resource identifier -> check resource/{resourceId}/operations and resource/{resourceId}/operations/{operationId}  exists

How does it sound?

julius-jogela: Sounds right — done in 938a514.

I had allowed a mutation of a single resource to satisfy the rule with the Operations endpoints at the *parent collection* as well, on the assumption that both levels were acceptable. That fallback is gone; the check is now exactly as you describe: the Operations endpoints must sit under the resource path of the mutation itself, collection or instance

Also rebased: #1448 has merged, so this is now a single commit on `main` and the stacked-PR note is dropped. Test updated to cover the parent-collection shape as a violation instead of a pass; full IPA suite green (861 tests).

tools/spectral/ipa/rulesets/IPA-132.yaml
```diff
@@ -420,3 +421,31 @@ rules:
...
+
+      ##### Implementation details
+      Rule checks for the following conditions:
+
+        - Applies to mutating methods that declare a 202 response; the
+          `x-xgen-long-running-operation` extension is derived from the 202 during the merge
+          step, so the rule behaves identically on merged and per-service specs
+        - Legacy operations that predate IPA-132 are excluded from validation, whether marked
+          `legacy: true` during the merge step or matched by the shared legacy operationId list
+        - Both the Operations collection (`.../operations`) and the single Operation endpoint
```
yelizhenden-mdb, line 438: Only Get and List methods must be defined. Tests follow this, however I couldn't see it in the rule implementation and we can update the documentation accordingly

julius-jogela: Good catch — you're right, that was only enforced by accident in the tests. Fixed in 7d2bb80.

The implementation checked that the two path *keys* existed and never looked at the methods they defined. Every test happened to write `get: {}` on those paths, so the suite read as if the requirement were enforced, but a spec defining `/resource/operations` with only `post` would have passed the rule.

The rule now requires the `get` method on both endpoints — List on the Operations collection, Get on the single Operation — and the message ends with "each with a get method". Two new invalid cases cover it: a collection defined with only `post`, and a single Operation defined with only `delete`. Both passed before, both flag now.

On the "only" half: I've left "no method other than `get`" to `xgen-IPA-132-operation-must-be-a-read-only-resource`, which already rejects it on Operations endpoints. So this rule requires `get` to be present and that one forbids everything else — they compose without overlapping. Happy to pull the exclusivity check in here too if you'd rather have it in one place.

Documentation updated in `IPA-132.yaml` with the new bullet, README regenerated from it. Suite green at 863 tests.

andreaangiolillo (approved):
LGTM. Thanks for all the great contributions during your internship! 💯 🫡

PR #1476, opened 2026-09-08
fix: remove dangling $refs after HiddenEnvsFilter deletes hidden schemas

maks-m-mongo-leaf:
https://jira.mongodb.org/browse/CLOUDP-443394

## What

When `HiddenEnvsFilter` deletes a schema from `components/schemas`, properties in other schemas that reference it can be left with a dangling `$ref`. The next step that loads the spec calls `ResolveRefsIn`, which crashes with `map key "<SchemaName>" not found`.

Two property shapes are affected:
1. **Bare `$ref`** — `{"\": "#/components/schemas/Hidden"}` with no annotation. After `json.Unmarshal`, `Value == nil`; the existing filter skips it.
2. **`allOf` wrapper** — `{"allOf": [{"\": "..."}]}` with no `x-xgen-hidden-env` on the property itself. After the ref is pruned the property becomes empty and is left behind.

**Root cause in MMS PR #184045:** `ApiSearchAutoScalingView` was annotated `x-xgen-hidden-env: stage,prod` at the class level, but the `autoScaling` field on `ApiSearchDeploymentRequestView` was added without the same annotation. Fixed in MMS by PR #185021 (backported to `v20260909` by PR #185860). This PR makes `foascli` resilient to the same class of mistake.

## Fix

Added `removeDanglingSchemaRefs` called after `applyOnSchemas`. It scans component schema properties and removes any that are a dangling bare `$ref` or an `allOf`/`anyOf`/`oneOf` wrapper that became empty after its refs were pruned.

## Testing

Table-driven test `TestApply_DanglingRefs` covers the three real-world property shapes found in the MMS OAS:
1. `allOf` wrapper without annotation on the property (the original bug)
2. `allOf` wrapper with `x-xgen-hidden-env` on the property (existing behaviour preserved)
3. Bare `$ref` without annotation

### DIFF
--- a/tools/foas/openapi/filter/hidden_envs.go
+++ b/tools/foas/openapi/filter/hidden_envs.go
@@ -57,7 +57,80 @@ func (f *HiddenEnvsFilter) Apply() error {
 		return nil
 	}
 
-	return f.applyOnSchemas(f.oas.Components.Schemas)
+	if err := f.applyOnSchemas(f.oas.Components.Schemas); err != nil {
+		return err
+	}
+
+	f.removeDanglingSchemaRefs(f.oas.Components.Schemas)
+	return nil
+}
+
+// removeDanglingSchemaRefs removes properties whose $ref (bare or via allOf/anyOf/oneOf)
+// points to a schema deleted by HiddenEnvsFilter.
+func (*HiddenEnvsFilter) removeDanglingSchemaRefs(schemas openapi3.Schemas) {
+	for _, schema := range schemas {
+		if schema == nil || schema.Value == nil {
+			continue
+		}
+		for propName, prop := range schema.Value.Properties {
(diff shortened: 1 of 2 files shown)


wtrocki:
Good catch on this use case. 
Wondering if we the tool can call generic kinapi/go openapi validation?
Alternatively we can fail when a reference is missing. 
If we normalize (pun intended) missing refs, we're de facto hiding the exact problem.

maks-m-mongo-leaf:
> Good catch on this use case. Wondering if we the tool can call generic kinapi/go openapi validation? Alternatively we can fail when a reference is missing. If we normalize (pun intended) missing refs, we're de facto hiding the exact problem.

I also fear to hide the problem, therefore I'd rather close the PR.

